Issue #194, opened 2022-10-01
Move logs into `LOCALAPPDATA/.ASLM/LOGS`, don't overwrite


zacsimile:
Rather than store the logs within the code base, store them locally on a per-machine basis in the `LOCALAPPDATA/.ASLM/LOGS` folder. This is a similar approach to storing the configuration files in `LOCALAPPDATA/.ASLM/config` (see `aslm.config.config.py`). 

Let's switch to a per-run logging mode, where we create new log files each time the program is launched. Each of these log files should be date and time stamped. Potentially we could associate a UUID with them as well, if necessary (date and time stamps are not unique).

PR #295, opened 2023-01-02
Logging Now in APPDATA

AdvancedImagingUTSW:
#194

I think this may be most of the way there already.  Now the logs are being saved into LOCALAPPDATA/.ASLM/LOGS into a sub-directory that specifies the YYYY-MM-DD-HHMM.

If @codeCollision4 can look over this and make sure everything is operating as expected, I think we are good to go.  Will initiate a PR.

Co-Authored-By: Kevin Dean <42547789+AdvancedImagingUTSW@users.noreply.github.com>

### DIFF
--- a/src/aslm/config/config.py
+++ b/src/aslm/config/config.py
@@ -1,3 +1,35 @@
+# Copyright (c) 2021-2022  The University of Texas Southwestern Medical Center.
+# All rights reserved.
+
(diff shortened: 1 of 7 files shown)


zacsimile (approved):
Looks good. Added one intermediate file path creation to initialize the `logs` folder.

Issue #360, opened 2023-02-08
Can't enter any number in the galvo frequency field


zacsimile:
All values are invalid, it seems.

PR #362, opened 2023-02-08
Galvo frequency

AdvancedImagingUTSW:
Fix for #360.  Separated into 2 separate commits.  One with the single line of code changed, and the second with commenting the code so that it is to our spec.


zacsimile (approved):
Works on multiscale.

PR #390, opened 2023-02-12
Add time series feature

zacsimile:
Lets us take a time series of single acquisitions.


annie-xd-wang (approved):
The code works. I'm just curious about the time series feature that takes "timepoints" frames in total. If we want to take "timepoints" frames for each channel, we could use nested loops.

Issue #398, opened 2023-02-22
Stack pause no longer implemented?


AdvancedImagingUTSW:
Trying to acquire a z stack on live cells rights now and the stack pause does not seem to work.  Looking at feature and I don't see it as well.

PR #399, opened 2023-02-24
stack pause feature

annie-xd-wang:
#398, add an independent stack pause feature and add it to the current ZStackAcquisition workflow.


zacsimile (approved):
Looks good and confirmed working on the multiscale.

PR #414, opened 2023-03-26
Waveform parameters

### LINKED ISSUE #413: Delay, smoothing, and duty cycle not implemented.
Waveform popup parameters not implemented.  

Spent some time working on this today.  See branch waveform_parameters

Made good progress. Somewhat close to a PR.

AdvancedImagingUTSW:
Meant for issue #413.

Ultimately, this is a partial PR. I want to check with @zacsimile before we go any further on how we want to implement certain aspects. As is, I do not think it should change operation so long as the smoothing percent value is at 0.


zacsimile (approved):
Appears to work on the multiscale.

annie-xd-wang (approved):

Issue #289, opened 2022-12-17
Eliminate initial stage positions


zacsimile:
Drove the stage right into the bottom of the chamber on startup. Nothing I could do about it.

AdvancedImagingUTSW:
Here is some code that is tested in the debugging mode of Pycharm.  I think if we combine with this the `None` referencing mode by default, it would be a good start.

```

def build_PIStage_connection(controller_name, serial_number, stages, reference_modes):
    """Connect to the Physik Instrumente Stage

    Parameters
    ----------
    controller_name : str
        Name of the controller, e.g., "C-863.11"
    serial_number : str
        Serial number of the controller, e.g., "0112345678"
    stages : str
        Stages to connect to, e.g., "M-111.1DG"
    reference_modes : str
        Reference modes for the stages, e.g., "FRF"

    Returns
    -------
    stage_connection : dict
        Dictionary containing the pi_tools and pi_device objects
    """
    pi_stages = stages.split()
    pi_reference_modes = reference_modes.split()
    pi_tools = pitools
    pi_device = GCSDevice(controller_name)
    pi_device.ConnectUSB(serialnum=serial_number)
    pi_tools.startup(
        pi_device, stages=list(pi_stages), refmodes=list(pi_reference_modes)
    )

    # wait until pi_device is ready
    block_flag = True
    while block_flag:
        if pi_device.IsControllerReady():
            block_flag = False
        else:
            time.sleep(0.1)

    # Are we on target? What is our reference mode state? Is the Servo on?
    on_target = pi_device.qFRF()
    servo_on = pi_device.qSVO()
    all_on_target = all(value for value in on_target.values())
    all_servo_on = all(value for value in servo_on.values())
    if all_servo_on and all_on_target:
        pass
    else:
        messagebox.showwarning(title="ASLM",
                               message="Physik Instrumente Stages Unreferenced or Not On Target. \n"
                                       "Proceed With Caution")

    stage_connection = {"pi_tools": pi_tools, "pi_device": pi_device}
    return stage_connection
```

PR #421, opened 2023-04-01
Map configuration.yaml axes onto PI stage axes in order they are passed

zacsimile:
Re: [#289#issuecomment-1493000371](https://github.com/TheDeanLab/ASLM/issues/289#issuecomment-1493000371)


AdvancedImagingUTSW (approved):
Very nice. We will have to add this to the documentation at some point.

PR #426, opened 2023-04-12
420 pressed stop and then acquire in too quick a succession program stopped working

### LINKED ISSUE #420: Pressed Stop and then Acquire in too quick a succession, program stopped working
Log attached. Traceback:

```
error message:  DCAMERR.TIMEOUT
camera thread exception happened! Two different threads tried to use the same ObjectInSubprocess at the same time! This is bad. Look at the docstring of concurrency_tools.py to see an example of how to use a _Custody object to avoid this problem. Traceback (most recent call last):
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\thread_pool.py", line 175, in func
    target(*args, **kwargs)
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\controller.py", line 957, in capture_image
    self.model.run_command(command)
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\model\concurrency\concurrency_tools.py", line 404, in __getattr__
    with self._.parent_pipe_lock:
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\model\concurrency\concurrency_tools.py", line 578, in __enter__
    raise RuntimeError(
RuntimeError: Two different threads tried to use the same ObjectInSubprocess at the same time! This is bad. Look at the docstring of concurrency_tools.py to see an example of how to use a _Custody object to avoid this problem.

camera thread ended because of exception!: not all arguments converted during string formatting
```
[model_debug.log](https://github.com/TheDeanLab/ASLM/files/11129852/model_debug.log)

annie-xd-wang:


zacsimile (approved):
Code looks good and it appears to work on the multiscale.

PR #427, opened 2023-04-14
metadata supports constant velocity acquisition mode

### LINKED ISSUE #425: Data_Couln't_be_Saved_Properly_In_ConstantVelocityAcquisition
After acquiring data over a certain range with ConstantVelocityAcquisition mode, ideally we should be getting a stack of data with 7 slices, but only 1 slice or only one frame has been saved. But in the debug log file, it shows multiple frames have been received. 

```
2023-04-06 17:02:33,538 - model - INFO - model: ASLM Model - Received frames in total: 36
```

annie-xd-wang:


AdvancedImagingUTSW (approved):
Simple enough. Thanks!

PR #431, opened 2023-04-20
428 resizable popup camera view window

### LINKED ISSUE #428: Resizable Popup Camera View Window
When the user right-clicks and pops the window out, @JinlongL and @zacsimile have expressed that it would be nice if we could make this window resizable. 

One option is to have not a purely dynamic resize, but have it grow by some integer value.  Or, if @annie-xd-wang is very creative, maybe it could resize completely as the user adjusts it. Only posting this hear because @annie-xd-wang volunteered herself for it.

annie-xd-wang:
The image will appear completely as a square on the screen. If we want the image to take up the entire screen (usually a rectangle), I need to add a scrollbar to the canvas. However, the image is only partially displayed unless the canvas is scrolled.

### DIFF
--- a/src/aslm/controller/controller.py
+++ b/src/aslm/controller/controller.py
@@ -1094,6 +1094,7 @@ class Controller:
                 camera_view_controller = CameraViewController(
                     popup_window.camera_view, self
                 )
+                popup_window.popup.bind("<Configure>", camera_view_controller.resize)
                 self.additional_microscopes[microscope_name][
                     "camera_view_controller"
                 ] = camera_view_controller
--- a/src/aslm/controller/sub_controllers/camera_view_controller.py
+++ b/src/aslm/controller/sub_controllers/camera_view_controller.py
@@ -1102,12 +1102,15 @@ class CameraViewController(GUIController):
     def refresh(self, width, height):
         if width == self.width and height == self.height:
             return
-        self.canvas_width = width - 151
-        self.canvas_height = height - 85
-        self.view.canvas.config(width=self.canvas_width, height=self.canvas_height)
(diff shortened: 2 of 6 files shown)


AdvancedImagingUTSW:
After quick glance @annie-xd-wang, it seems as if a test has a hard-coded window size of 512x512, which is probably no longer valid for the larger window...

zacsimile:
On synthetic hardware mode on Mac the window keeps expanding on its own. This is if I just open the program and press nothing.

1 second:
image

5 seconds:
image

10 seconds:
image

AdvancedImagingUTSW:
Although I didn't notice this behavior this first time, I can confirm that it is happening on my machine too. However, it only happens if you do not resize the window, if that gives a clue. Once I pop the camera view window out, and adjust the window, it remains stationary. Otherwise, it grows indefinitely (both popped in and out).

zacsimile:
Confirmed working on my Mac now.

annie-xd-wang:
It doesn't happen on my machine which is Windows. I tested it on a Mac machine today, it does expand on its own when launching the software. I've uploaded a fix. However, that Mac is an old one and doesn't support popping up the camera view window. Please confirm if anything wrong when popping up the window.

zacsimile:
For me it now doesn't expand on its own when I launch and the resizing works if I pop the window out. The window barely fits on my screen if it's not popped out so I'm not sure if it's working a non-popped out window. But I think it does work.

annie-xd-wang:
I only add resizing event binding to the popup windows (both the main camera viewer and additional camera viewer).

zacsimile (approved):

PR #432, opened 2023-04-20
157 robust peak detection autofocus

### LINKED ISSUE #157: Robust Peak Detection - Autofocus
We should implement a method that fits the autofocus data with a Gaussian, as well as some other metrics, to make sure that we aren’t just finding the max value in a noisy signal.

AdvancedImagingUTSW:
Still need to test this on the machine...

### DIFF
--- a/src/aslm/config/experiment.yml
+++ b/src/aslm/config/experiment.yml
@@ -35,6 +35,7 @@ AutoFocusParameters:
   fine_range: 50
   fine_step_size: 5
   fine_selected: True
+  robust_fit: True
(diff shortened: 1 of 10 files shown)


AdvancedImagingUTSW:
I was hoping that @annie-xd-wang could review the code, and once approved, then @zacsimile would test the autofocus routine on the microscope. 

Sound good?  

After mucking around with some of the menu items, it almost seems like this could be another sub-controller...  Just a thought.

annie-xd-wang (approved):

PR #448, opened 2023-05-12
calculate focus ramp for a Z-Stack

annie-xd-wang:


src/aslm/model/metadata_sources/metadata.py
```diff
@@ -149,16 +156,22 @@ def set_shape_from_configuration_experiment(self) -> None:
...
+        #         self.configuration["experiment"]["MultiPositions"]["stage_positions"]
+        #     )
+        # else:
+        #     self.positions = 1
+        
+        # let the data sources have the ability to save more frames
+        self._multiposition = True
+        self.positions = len(
+            self.configuration["experiment"]["MultiPositions"]["stage_positions"]
+        ) * 50
```
zacsimile: This `*50` feels really sketchy. Can we justify 50 times the number of positions by default?

zacsimile: Could we have certain features (e.g. `ZStack`) set the image writer's `_multiposition` flag to `True` instead of leaving it on by default?

zacsimile:
It seems the display no longer updates as images are acquired.

zacsimile:
The program also hangs indefinitely when pressing stop, even though the stop command is logged in `model_debug.log`.
[model_debug.log](https://github.com/TheDeanLab/ASLM/files/11546052/model_debug.log). This may be related to #450.

src/aslm/model/metadata_sources/metadata.py
annie-xd-wang: Yeah, *50 is quite sketchy. Adding such a value just to make sure the global image writer won't throw any error when there are a lot of images comes back other than ZStack, such as Autofocus. Actually, I think the metadata should have a mode to support unlimited image saving instead of calculate exactly how many frames should be saved. Adding _multiposition flag to a feature is simple, but it won't solve the global image writer error. We can look into the metadata to see what's a better way to implement it later.

zacsimile (approved):

PR #452, opened 2023-05-18
451 abstract autofocus feature

### LINKED ISSUE #451: Abstract Autofocus Feature
We need to create a more abstract autofocus feature. The new autofocus feature can move a separate stage axis, and it can also move the offset of a NI remote focus device.

annie-xd-wang:


zacsimile:
This is super cool and it seems to work! The only issue is we need to be able to pass a floating point range/step size in for remote_focus and the values need to be able to go negative.

image


It would also be nice if by default this launches with device `stage` and device reference `focus` by default.

AdvancedImagingUTSW:
We should be able to test this out on CT-ASLM-V1.

AdvancedImagingUTSW:
Confirmed working with changes in #503.

AdvancedImagingUTSW (approved):

zacsimile (approved):

Issue #447, opened 2023-05-11
Some odd default values experienced


AdvancedImagingUTSW:
While setting up CT-ASLM-V1 to run on this software, I noticed a few somewhat odd pieces of behavior. Nothing major, and I think I know the source of it.  

Let's say I change the filter names in the configuration file. Well, when I open the software, it uses the experiment file to load the default settings in the view, and now, these don't match what is in the configuration. It almost seems like we should overwrite the experiment file in the .config folder every time we close the software.  This way, when you open it, it will load what you used last time, not some odd default behavior that may not be relevant...

AdvancedImagingUTSW:
I will provide an up-to-date assessment of this branch, which I need some help with shortly.

AdvancedImagingUTSW:
@annie-xd-wang, @zacsimile. 

This branch is a bit confusing since I made a mistake with git. Essentially, I updated the controller so that it automatically saves over the experiment.yml file upon closing the file. All fine, Zach also agrees that this seems like a good idea.  

However, after I did that, I merged it into another branch where I was working on the ASI filter wheels. As a result of a mistake, all of those changes are now in this branch...  Anyhow, we will just make a single pull request, and we are getting close to being ready.  But, there is one thing that I wanted to discuss. 

We have a few instances where multiple devices share a common API/communication method. For example, the NIGalvoStage, lasers, and shutters, all use the NIDAQ device. Consequently, we have to pass a single `device_connection` to each of these in the `microscope` object. Currently, most of these are kinda hard-coded in.

Would it make more sense to complicate our configuration.yml file a bit more to distinguish hardware that is shared versus those that are not?  For example, something like this?

```
shared_hardware:
  -
      type: ASIFilterWheel and ASIStage
      port: COM19
      baudrate: 9600
  -
      type: NI and NILaser and NIGalvo
      port: ...
      baudrate: ...
hardware:
  daq:
    type: NI
  camera:
    -
      type: HamamatsuOrca
      serial_number: 302158
    -
      type: HamamatsuOrca
      serial_number: 302352
  filter_wheel:
    type: SutterFilterWheel
    port: COM19
    baudrate: 9600
    number_of_wheels: 2
```

This could make the microscope object a little less kludged?

zacsimile:
Sorry, it's been a minute since I looked deep into this. Can you provide an example or two of the kludges this would avoid?

AdvancedImagingUTSW:
In `microscope.py`

line 179:
```
[...]

Issue #429, opened 2023-04-17
ASI Filter Wheel


AdvancedImagingUTSW:
1.) Modify device startup functions so that only one connection object to the Tiger Controller is passed to the filter wheel and stages.

2.) Also need to add the filter wheel command set to the `asi_tiger_controller` API. Commands can be found [here](http://asiimaging.com/docs/fw_1000).

At some point, would also be valuable to discuss how to independently handle each filter wheel. But for now, let's put that on the back burner and focus on points 1 and 2 above.  These are most necessary in the short-term for @JinlongL

AdvancedImagingUTSW:
Merged with #457

PR #454, opened 2023-05-19
ASI Filter Wheel, Save GUI Settings on Exit, and Microscope Object

AdvancedImagingUTSW:
Addresses:
#447 - Added functionality to the controller that now overwrites the .config/experiment.yml file upon exit. Thus, when the program is reloaded, it is in the same state that it was exited in. 

#429 - ASI Filter Wheel Commands now added. Only one instance of the device_connection can exist for the ASI Tiger Controller, which controls both the filter wheels, dichroic changers (not yet implemented), and stages.  Cleaned up the Microscope object to make the dictionary parsing and logic slightly more clear (broke it up into a few more sub functions). I believe the logic now makes sense to me, but I will need to make sure that the ASI device_connection gets passed to both the stages and filter wheels in the morning. 

Also, given that I mucked around with the microscope object, which is obviously critical, we should test this on more than one microscope.

### DIFF
--- a/src/aslm/model/device_startup_functions.py
+++ b/src/aslm/model/device_startup_functions.py
@@ -355,37 +355,30 @@ def start_stage(
 
     if device_type == "PI":
         from aslm.model.devices.stages.stage_pi import PIStage
-
         return PIStage(microscope_name, device_connection, configuration, id)
 
     elif device_type == "MP285":
         from aslm.model.devices.stages.stage_sutter import SutterStage
-
         return SutterStage(microscope_name, device_connection, configuration, id)
 
     elif device_type == "Thorlabs":
         from aslm.model.devices.stages.stage_tl_kcube_inertial import TLKIMStage
-
         return TLKIMStage(microscope_name, device_connection, configuration, id)
 
     elif device_type == "MCL":
         from aslm.model.devices.stages.stage_mcl import MCLStage
-
         return MCLStage(microscope_name, device_connection, configuration, id)
 
     elif device_type == "ASI":
         from aslm.model.devices.stages.stage_asi import ASIStage
-
         return ASIStage(microscope_name, device_connection, configuration, id)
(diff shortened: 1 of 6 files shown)


src/aslm/model/device_startup_functions.py
```diff
@@ -287,19 +287,23 @@ def load_stages(configuration, is_synthetic=False):
...
-
-            stage_devices.append(
-                auto_redial(
-                    build_ASI_Stage_connection,
-                    (
-                        stage_config["port"],
-                        stage_config["baudrate"],
-                    ),
-                    exception=TigerException,
+            filter_wheel = configuration["configuration"]["hardware"]["filter_wheel"]['type']
```
annie-xd-wang, line 290: This is assuming that if there is an ASI filter wheel, the filter wheel will be the same hardware of an ASI stage. It works, but I think we'd better refactor current device loading structure to make sure devices connect to the right hardware without hard coding.

annie-xd-wang (approved):
I've read the code, but have the feeling that we should refactor the devices loading structure. Should we refactor before merging this branch? or create a new issue and refactor later?

src/aslm/model/device_startup_functions.py
AdvancedImagingUTSW: I think it is quite difficult to account for every single permutation. Technically, there could be three different scenarios for an ASI filter wheel, and they would have to be handled differently.

ASI offers an MS2000 controller, a Tiger Controller, and an independent Filter Wheel controller. Some of these have different command sets. So in the meantime, we will have to stick with a Tiger Controller.

AdvancedImagingUTSW:
I'm somewhat tempted to create an issue that lists various things that should be refactored. But then again, I understand that if it doesn't happen soon, it probably won't happen.

How long do you think it would take to refactor it?

PR #464, opened 2023-05-26
459 refactor stages

### LINKED ISSUE #459: Refactor Stages
@annie-xd-wang has wanted to streamline the stage control. Currently, makes a lot of repetitive calls to the controller.

annie-xd-wang:
Need to test with stage devices. Will add more test cases.


AdvancedImagingUTSW:
Let's schedule times to test with each of the physical devices.

MP-285 - @AdvancedImagingUTSW 
PI - @zacsimile 
ASI - @JinlongL 
GalvoStage - @BingyingChen

zacsimile:
It doesn't have to be @annie-xd-wang, but can someone write how `axes_mapping` in `configuration.yaml` works into our documentation  before merging this PR, so we don't forget? 

My understanding is as follows.

* `axes_mapping` lists the values we use to access stage axes on the stage controller hardware.
* `axes` tells us which software axis (X, Y, Z, Theta or Focus) is controlled by the hardware stage axis at the corresponding index of `axes_mapping`.
* PI stages will tell you their `axes_mapping` in a consistent fashion if queried from their API, so we do not need to specify this information for a PI stage in `configuration.yaml`. Instead, we assume the user is aware of this mapping in advance (e.g., because they looked at PIMikroMove) and has passed the stage axes in `axes` in the order they will be found on the hardware controller.
* ASI stages do not provide their `axes_mapping` via their API, and so we do need to specify this information for ASI stages in `configuration.yaml`.

AdvancedImagingUTSW:
I can write some documentation

zacsimile:
Went to benchmark timing on this and it turns out that z-stack runs continuously.
[model_debug.log](https://github.com/TheDeanLab/ASLM/files/11710592/model_debug.log)

AdvancedImagingUTSW:
What do you mean?

zacsimile:
I mean Z-stacking never ends. If you select Z-stack and hit go, it runs an infinite loop of z-stacks as if you are in continuous mode. Checked out d88e0208a92a63e74e939e6f1deb79711099c8d9 to make sure it wasn't my edits (which have a test failure I'm currently fixing) and the problem is there as well.

zacsimile:
Like so

https://github.com/TheDeanLab/ASLM/assets/1263313/7f140d18-b46e-470d-9996-3651a452a47f

AdvancedImagingUTSW:
On the plus side it looks faster

annie-xd-wang:
The timepoints is 100, which might be the reason that ZStack runs again and again.

zacsimile:
You are right. It was the timepoints. I didn't notice they were still set at 100.

AdvancedImagingUTSW:
Final task. Test on @BingyingChen's microscope to confirm that everything works, then merge.

AdvancedImagingUTSW:
@zacsimile - I put some minimal documentation on how to map the axes here: https://thedeanlab.github.io/ASLM/software_configuration.html#stage-axes-definition

annie-xd-wang:
Just tested with @BingyingChen on her microscope, this branch works on her device.

zacsimile (approved):

PR #475, opened 2023-06-05
Add overlap to volume search tiles

zacsimile:


AdvancedImagingUTSW:
@zacsimile, why did codecov/project fail? Is it because the threshold of unit tests fail? I didn't see the threshold defined in the codecov configuration as shown [here](https://docs.codecov.com/docs/commit-status)...

PR #476, opened 2023-06-05
acquire all the selected channel in Single Acquisition

### LINKED ISSUE #474: Single acquisition only acquires first selected channel, rather than all selected channels

annie-xd-wang:


AdvancedImagingUTSW:
@annie-xd-wang - Is there anyway we can get a unit test for this PR?

zacsimile:
@annie-xd-wang Seconded. It would be ideal if we had a unit test to confirm that single acquisition is spitting out the correct number of channels to alert us to this kind of issue in the future.

zacsimile (approved):

PR #477, opened 2023-06-06
Device mock test

AdvancedImagingUTSW:
Wanted to learn a bit more about mocking devices, so I used this as an excuse. Tests the Sutter filter wheel class. 

The next thing I'm hoping to figure out is patching objects. In principle, it could be nice to patch the `device_connection` object with a mock device connection, thereby allowing us to run the Sutter filter wheel class in a non-test environment but not actually connected to the hardware.

Let me know if you think this is valuable.  If so, I don't mind doing a couple of other pieces of hardware.


AdvancedImagingUTSW (commented):
I looked over these again. 10 files changed, 8 of which were the documentation, 1 was the tests for the filter wheel, and 1 was the filter wheel itself. Confirmed that all changes to the filter wheel should be fine.

annie-xd-wang (approved):

PR #479, opened 2023-06-09
Clean up camera configuration section, implement getters, and retrieve values from device.

AdvancedImagingUTSW:
- Provide default values for many camera parameters should the user not provide them.
- Add ability to get the camera pixel size from the API.
- Eliminate the manual calculation of the readout time. Now pulls actual value from the API, which eliminates camera model-specific mismatches.


AdvancedImagingUTSW:
Ugh. Ugly.

PR #481, opened 2023-06-14
Subclass Hamamatsu ORCA-Lightning

zacsimile:
The lightning is sufficiently different to need changes in the light sheet readout parameters.

Here I also incorporate @annie-xd-wang 's solution in de8b3f4514688fe8427c76caf7a1a8269932c6aa / #409 as hardcoded. This may throw a warning in `set_property_value()`, but it effectively always sets the camera to run in its fastest mode. I also kill readout speed in the Lightning `__init__()`, since this property is not used.


zacsimile:
@JinlongL please try this for your system.

annie-xd-wang (approved):

PR #491, opened 2023-06-19
Circumvent error arising from calling GUI function outside of GUI thread

zacsimile:
The `update_event` thread makes GUI calls outside of main Tk loop, This results in 

```
  File "C:\Users\DeanLab\.conda\envs\ASLM\lib\threading.py", line 973, in _bootstrap_inner
    self.run()
  File "C:\Users\DeanLab\.conda\envs\ASLM\lib\threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\controller.py", line 953, in update_event
    self.update_stage_controller_silent(value)
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\controller.py", line 918, in update_stage_controller_silent
    self.stage_controller.set_position_silent(stage_gui_dict)
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\sub_controllers\stage_controller.py", line 306, in set_position_silent
    self.unbind_position_callbacks()
  File "c:\users\deanlab\desktop\github\aslm\src\aslm\controller\sub_controllers\stage_controller.py", line 250, in unbind_position_callbacks
    self.widget_vals[axis].trace_remove("write", cbname)
  File "C:\Users\DeanLab\.conda\envs\ASLM\lib\tkinter\__init__.py", line 424, in trace_remove
    self._tk.call('trace', 'remove', 'variable',
RuntimeError: main thread is not in main loop
```

from time to time. The right solution is to move all GUI calls out of external threads. The solution I propose here works around the threading so we have to change less code.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,7 @@
+coverage:
+  status:
+    project: off
+    patch: off
 ignore:
   - "test"
   - "design"
--- a/src/aslm/controller/controller.py
+++ b/src/aslm/controller/controller.py
@@ -39,6 +39,7 @@ import multiprocessing as mp
 import threading
 import sys
 import os
+import time
 
 # Third Party Imports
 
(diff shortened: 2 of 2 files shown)


AdvancedImagingUTSW:
“  # ZM: I am so sorry for this.”
😂

AdvancedImagingUTSW (approved):
What would happen in the case that it tries to update the stage controller silently 10 times, and it throws a RuntimeError each time?

zacsimile:
The stage controller wouldn't update. So we'd have a mismatch between true stage position and what is reported in the program.

PR #496, opened 2023-06-28
Update laser_ni.py

AdvancedImagingUTSW:
For the PCIE6738 card, we have an abundance of analog outputs, but not many digital outputs. So, we need to sometimes make an analog output act like a digital output. I have added some logic to make this possible. Works on the SpectralTIRF system.

### DIFF
--- a/src/aslm/model/devices/lasers/laser_ni.py
+++ b/src/aslm/model/devices/lasers/laser_ni.py
@@ -56,18 +56,19 @@ class LaserNI(LaserBase):
             self.laser_min_do = self.device_config["onoff"]["hardware"]["min"]
             self.laser_max_do = self.device_config["onoff"]["hardware"]["max"]
             self.laser_do_task = nidaqmx.Task()
-            try:
-                # Digital Modulation via a Digital Port
-                self.laser_do_task.do_channels.add_do_chan(
-                    laser_do_port, line_grouping=LineGrouping.CHAN_FOR_ALL_LINES)
-                self.on_off_type = "digital"
-            except DaqError as e:
+
+            if "/ao" in laser_do_port:
                 # Artificial Digital Modulation via an Analog Port
                 self.laser_do_task.ao_channels.add_ao_voltage_chan(
                     laser_do_port, min_val=self.laser_min_do, max_val=self.laser_max_do)
                 self.on_off_type = "analog"
 
-        except KeyError as e:
+            else:
+                # Digital Modulation via a Digital Port
+                self.laser_do_task.do_channels.add_do_chan(
+                    laser_do_port, line_grouping=LineGrouping.CHAN_FOR_ALL_LINES)
+                self.on_off_type = "digital"
+        except (KeyError, DaqError) as e:
             self.laser_do_task = None
             if isinstance(e, DaqError):
                 logger.exception(e)
--- a/src/aslm/model/devices/lasers/laser_ni.py
+++ b/src/aslm/model/devices/lasers/laser_ni.py
@@ -48,16 +48,26 @@ logger = logging.getLogger(p)
 class LaserNI(LaserBase):
     def __init__(self, microscope_name, device_connection, configuration, laser_id):
         super().__init__(microscope_name, device_connection, configuration, laser_id)
+        self.on_off_type = None
 
         # Digital out (if using mixed modulation mode)
         try:
             laser_do_port = self.device_config["onoff"]["hardware"]["channel"]
-
+            self.laser_min_do = self.device_config["onoff"]["hardware"]["min"]
+            self.laser_max_do = self.device_config["onoff"]["hardware"]["max"]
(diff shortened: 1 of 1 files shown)


src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -48,16 +48,26 @@
...
+                self.on_off_type = "digital"
+            except DaqError as e:
+                # Artificial Digital Modulation via an Analog Port
+                self.laser_do_task.ao_channels.add_ao_voltage_chan(
+                    laser_do_port, min_val=self.laser_min_do, max_val=self.laser_max_do)
+                self.on_off_type = "analog"
+
+        except KeyError as e:
             self.laser_do_task = None
             if isinstance(e, DaqError):
```
zacsimile, line 73: This check will always be `False` since we are excepting a `KeyError`.

src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -48,16 +48,26 @@
 class LaserNI(LaserBase):
     def __init__(self, microscope_name, device_connection, configuration, laser_id):
         super().__init__(microscope_name, device_connection, configuration, laser_id)
+        self.on_off_type = None
 
         # Digital out (if using mixed modulation mode)
         try:
```
zacsimile, line 54: Try/excepting ao vs do feels dangerous. Could we create two classes of LaserNI instead, where we can explicitly specify which is which from the YAML file?

src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -48,16 +48,26 @@
...
             laser_do_port = self.device_config["onoff"]["hardware"]["channel"]
-
+            self.laser_min_do = self.device_config["onoff"]["hardware"]["min"]
+            self.laser_max_do = self.device_config["onoff"]["hardware"]["max"]
             self.laser_do_task = nidaqmx.Task()
-            self.laser_do_task.do_channels.add_do_chan(
-                laser_do_port, line_grouping=LineGrouping.CHAN_FOR_ALL_LINES
-            )
-        except (KeyError, DaqError) as e:
+            try:
```
zacsimile: Try/excepting ao vs do feels dangerous. Could we create two classes of LaserNI instead, where we can explicitly specify which is which from the YAML file?

AdvancedImagingUTSW:
Updated to search the port name to identify if we are using an analog or a digital port.

src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -48,15 +48,26 @@
 class LaserNI(LaserBase):
     def __init__(self, microscope_name, device_connection, configuration, laser_id):
         super().__init__(microscope_name, device_connection, configuration, laser_id)
+        self.on_off_type = None
 
         # Digital out (if using mixed modulation mode)
         try:
             laser_do_port = self.device_config["onoff"]["hardware"]["channel"]
-
+            self.laser_min_do = self.device_config["onoff"]["hardware"]["min"]
```
zacsimile, line 56: Since we do not need this information for digital channels, I think we should move this into the `if "/ao"` control flow or make this a get or default.

src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -48,15 +48,26 @@
 class LaserNI(LaserBase):
     def __init__(self, microscope_name, device_connection, configuration, laser_id):
         super().__init__(microscope_name, device_connection, configuration, laser_id)
+        self.on_off_type = None
```
zacsimile, line 51: What about changing this to `self.on_off_type = "analog" if "/ao" in laser_do_port else "digital"` and then using this flag to make the decisions? Better yet, `self.on_off_analog = True if "/ao" in laser_do_port else False` and then the rest of the checks are just against a `bool` instead of a full string comparison (faster).

src/aslm/model/devices/lasers/laser_ni.py
```diff
@@ -98,7 +108,10 @@ def set_power(self, laser_intensity):
     def turn_on(self):
         try:
             if self.laser_do_task is not None:
-                self.laser_do_task.write(True, auto_start=True)
+                if self.on_off_type == "digital":
+                    self.laser_do_task.write(True, auto_start=True)
+                elif self.on_off_type == "analog":
+                    self.laser_do_task.write(self.laser_max_do, auto_start=True)
```
zacsimile, line 114: If you make `self.laser_max_do = self.device_config["onoff"]["hardware"].get("max", 1)` and `self.laser_min_do = self.device_config["onoff"]["hardware"].get("min", 0)` you can use Python's duck typing to automatically write `True` (`1`)/`False` (`0`) in the absence of specified min and max values and skip the `if self.on_off_type` checks entirely.

zacsimile (approved):

PR #499, opened 2023-07-04
498 fix hot keys

### LINKED ISSUE #498: Fix Hot Keys
Control + A should not be autofocus, we will have to designate as something else.

When the save popup dialog is up, keyboard hotkeys should be disabled.

Anytime we are dealing with a text box, no w, a, s, d...

AdvancedImagingUTSW:
Fix for the hot keys. Thought this could be a good opportunity for @mehr0096 to review the changes with @zacsimile, and to only merge if both agree that it looks okay.


mehr0096:
Hello Kevin, the stage control commands look to be disabled when a text box is displayed but I am not able to see a textbox to save any acquired data  I can save experiments though. I will go through the code review with Zach though to see if this an issue on my end.

AdvancedImagingUTSW:
There were some merge conflicts, so I had to update from develop.

zacsimile (approved):

PR #503, opened 2023-07-11
Effectively revert #502 to fix autofocusing

AdvancedImagingUTSW:
The #502 merge killed autofocusing. This restores it. We'll need to tweak how #502 works a bit.


annie-xd-wang:
I don't think #502 killed autofocusing. We need to set the stop_acquisiton_flag to False before running Autofocus.

annie-xd-wang (approved):

zacsimile:
You are right, this worked.

PR #505, opened 2023-07-11
Add photometrics camera

JinlongL:


zacsimile:
Confirmed tested on Jinlong's scope.

AdvancedImagingUTSW (approved):
So long as any problems are isolated to @JinlongL or Reto's microscopes, we can deal with them there.

AdvancedImagingUTSW:
Interestingly, the pvcam python module comes with unit tests, which are being run by GitHub. If we flag these as hardware, we should probably be okay.

zacsimile:
Good call. Updated.

PR #506, opened 2023-07-12
Extend tunable lens ramp by percent smoothed

zacsimile:
image

### DIFF
--- a/src/aslm/model/waveforms.py
+++ b/src/aslm/model/waveforms.py
@@ -388,7 +388,7 @@ def smooth_waveform(waveform, percent_smoothing=10):
 
     """
     waveform_length = np.size(waveform)
-    window_length = int(np.floor(waveform_length * percent_smoothing / 100))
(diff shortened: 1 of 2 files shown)


zacsimile:
Likely requires an update to `self.sweep_time` in `daq_base.py`. This corresponds to a necessary change in `daq_ni.py`'s `self.camera_low_time` to modify the duty cycle. *However*, the synthetic mode tests indicate that this isn't necessary, so I will try it on a real machine before merging.

Update would be something like,

```
# Smoothing slightly increases the duration of the tunable lens ramp
                ps = float(self.waveform_constants["remote_focus_constants"][microscope_name][self.zoom][channel["laser"]].get("percent_smoothing", 0.0))
                if ps > 0:
                    self.sweep_time += ps/100
```

where `self.zoom = self.configuration["experiment"]["MicroscopeState"]["zoom"]`.

annie-xd-wang (approved):

PR #508, opened 2023-07-13
Add proper sweep_time calc to smoothing

zacsimile:
Finishes #506. Moves sweep time calculation into single spot in `microscope.py`. Makes sure the waveforms all properly adjust on change of percent smoothing of remote focus ramp.


zacsimile:
Confirmed working on the ctASLMv1 with the oscilloscope.

annie-xd-wang (approved):

PR #522, opened 2023-07-14
Update microscope.py

### LINKED ISSUE #516: Force position request at the start of acquisition
In case the stage has become out-of-whack with what our program positions report (e.g. from using a joystick for movement), we should hard request the stage position once at the start of `microscope.py`'s `prepare_acquisition()`.

AdvancedImagingUTSW:


AdvancedImagingUTSW:
Ah, OK. This works, but we actually need to do this in the controller in order to get the positions to update in the GUI.

zacsimile:
Seems good to me! Works on the v1. Should we also add `self.active_microscope.ask_stage_for_position = True` above line 617 in the `"stop"` section of execute?

zacsimile (approved):

annie-xd-wang (approved):

PR #523, opened 2023-07-14
Slight safer safe write

zacsimile:
Thanks for the suggestion, @annie-xd-wang !

### DIFF
--- a/src/aslm/model/devices/APIs/sutter/MP285.py
+++ b/src/aslm/model/devices/APIs/sutter/MP285.py
@@ -80,7 +80,7 @@ class MP285:
         self.wait_until_done = True
 
         self.wait_time = 0.002
-        self.n_waits = int(timeout / self.wait_time)
+        self.n_waits = max(int(timeout / self.wait_time), 1)
 
         # Thread blocking here to prevent calls to get_current_position()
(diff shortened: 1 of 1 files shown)


src/aslm/model/devices/APIs/sutter/MP285.py
```diff
@@ -74,13 +74,13 @@ def __init__(self, com_port, baud_rate, timeout=0.25):
...
         self.serial.rtscts = True
-        
+
         self.speed = 1000  # None
         self.resolution = "high"  # None
         self.wait_until_done = True
 
         self.wait_time = 0.002
-        self.n_waits = int(timeout/self.wait_time)
+        self.n_waits = int(timeout / self.wait_time)
```
annie-xd-wang: We should make sure self.n_waits bigger than 0.

annie-xd-wang (approved):
Seems good to me except for making sure self.n_waits >= 1.

src/aslm/model/devices/APIs/sutter/MP285.py
zacsimile: Agree, fixed.

annie-xd-wang (approved):

PR #527, opened 2023-07-17
fix saving with customized features

annie-xd-wang:
closing the saving dialog window before confirming feature parameters.


zacsimile:
Confirmed working on the multiscale.

zacsimile (approved):

PR #532, opened 2023-07-19
Fix Ctrl + Mouse bindings

### LINKED ISSUE #531: Lambda callback issue on Control + Left/Right/Middle Mouse Click
```
  File "C:\Users\DeanLab\.conda\envs\ASLM\lib\tkinter\__init__.py", line 1892, in __call__
    return self.func(*args)
TypeError: <lambda>() takes 0 positional arguments but 1 was given
```

zacsimile:
Closes #531. `lambda: self.switch_tabs(1)` had no arguments, so one solution would be to pass it an argument. However, `<Control-1>` is Control + Left Mouse Click, and the tab switching is supposed to be Control + 1. Switching to `<Control-Key-1>` solves this.


AdvancedImagingUTSW:
To my understanding, <Button-1> is left click, and <Button-3> is right click. 
https://stackoverflow.com/questions/32289175/list-of-all-tkinter-events

AdvancedImagingUTSW:
At least testing on a Mac, it does throw the standard lambda error still despite switching tabs.
```
Exception in Tkinter callback
  File "/opt/miniconda3/envs/ASLM/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
    return self.func(*args)
TypeError: <lambda>() takes 0 positional arguments but 1 was given
```

This is because Tkinter passes an event along with the trace. Pushing one change so that  it does...

```
                    lambda event: self.switch_tabs(4),
```

AdvancedImagingUTSW (approved):

zacsimile:
Ah, good call. I'd missed that. Thanks for the fix!

PR #533, opened 2023-07-19
display the latest frame

annie-xd-wang:
The GUI will display the latest frame from the camera if it has completed showing one.

### DIFF
--- a/src/aslm/controller/controller.py
+++ b/src/aslm/controller/controller.py
@@ -676,6 +676,7 @@ class Controller:
                 and self.ilastik_controller.show_segmentation_flag
             )
 
+            self.stop_acquisition_flag = False
             self.launch_additional_microscopes()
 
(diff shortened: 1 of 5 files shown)


zacsimile:
This works well except if I have more than one channel selected and they have different exposure times. 

[model_debug.log](https://github.com/TheDeanLab/ASLM/files/12114865/model_debug.log)

annie-xd-wang:
The errors shown in the log file are not caused by the changes in this PR. Does the same error happen when you run multiple channels with the develop branch? I can look into the error when working with a device tomorrow.

zacsimile:
You are right, the same error occurs on develop. I hadn't tried this in a while. I will make a new issue.

zacsimile:
This is actually a known error I thought I'd addressed in #508.

zacsimile (approved):

PR #537, opened 2023-07-21
Reflect multiposition state changes in experiment config in controller

zacsimile:
This addresses a long-standing issue @AdvancedImagingUTSW pointed out, which is that the `is_multiposition` flag can be set to `True` without this being reflected in the GUI. The multiposition state that is pulled from the `experiment.yml` is now properly displayed in the GUI so the user sees the true state. We also update the multiposition table population to properly update the experimental state (previously it only updated the GUI state).


zacsimile:
Also a random f-string fix.

AdvancedImagingUTSW (approved):
Sorry, I should have reviewed this earlier...

Issue #473, opened 2023-06-03
ZStack `per_stack` and `per_z` return different channel orderings


zacsimile:
Strangely, `per_z` and `per_stack` seem to have different channel orders. For the same configuration in the GUI, Channel 2 on `per_stack` may be Channel 0 or 1 when `per_z` is selected.

<img width="452" alt="image" src="https://github.com/TheDeanLab/ASLM/assets/1263313/ef490fd1-3aa8-438b-a9fa-1f6db2f74c9e">

<img width="719" alt="image" src="https://github.com/TheDeanLab/ASLM/assets/1263313/83847bcb-d7ba-4b7b-8a35-148c8fc877ba">

AdvancedImagingUTSW:
@zacsimile - Annie is wondering if this is solved. If so, perhaps she can write a few unit tests to lock in the behavior.

zacsimile:
It is not solved. If @annie-xd-wang  could solve it and write unit tests, that would be awesome.

zacsimile:
For example, on the v2, per-stack saves in the correct order of (channel 0, channel 1, channel 2), but per z saves in the order of (channel 1, channel 2, channel 0). This does appear to be a saving issue, rather than an acquisition order issue.

Additionally, in per-stack (but not per-z), the final frame of the z-stack is in the first position of the stack. This is starting to feel like an issue with `_cztp_indices()`.

![image](https://github.com/TheDeanLab/ASLM/assets/1263313/2741a4e8-1899-4583-8542-6501bd24c5e4)

PR #538, opened 2023-07-21
fix ZStack perZ

annie-xd-wang:


zacsimile:
Confirmed on the ctASLMv2. Closes #473. @annie-xd-wang, please add a unit test that ensures this behavior for per-z/per-stack and then we'll merge.

zacsimile (approved):

PR #539, opened 2023-07-21
Add enabledisable joystick button

### LINKED ISSUE #517: Add enable/disable joystick button
The Sutter stages all come with joysticks that move the stage independently of our software. This currently forces a user to press the STOP stage button to alert our program to the new stage positions prior to moving the stage from within our program. This can be annoying. Instead, I propose we have a "Activate Joystick" that disables software control of the stage when selected and returns software control to the stage when de-selected. On de-selection, we should make a forced call for the new position from the stage.

3vwylie:
Joystick mode button implemented. It should appear only if joystick axes are specified in the configuration file. When pressed, it should disable the position entries and buttons corresponding to the specified joystick axes. Pressing the button should also update several visuals and hover texts corresponding to the joystick axes. Control + J should do the same thing as the button upon release. Switching microscopes should re-enable all buttons and turn "joystick mode" off, unless the joystick axes for both the current and former microscopes are the same, and joystick mode was on while the microscope switch was performed, in which case joystick mode will remain on. If switching from a microscope that has joystick axes and a microscope that does not, the button should disappear. The configuration file that comes with ASLM (C:\...\ASLM\src\aslm\config\configuration.yaml) has the default joystick axes (joystick_axes) set to x, y, and z, for both the meso- and nanoscale microscopes.


AdvancedImagingUTSW:
@3vwylie - We had a test fail when we made the PR. You can see where the problem arose by clicking details above in the tests section...  We will need to fix these before we can properly review everything...

```
================================== FAILURES ===================================
__________________ TestConfiguration.test_microscope_section __________________

self = <test_configuration.TestConfiguration testMethod=test_microscope_section>

    def test_microscope_section(self):
        expected_hardware = [
            "daq",
            "camera",
            "remote_focus_device",
            "galvo",
            "shutter",
            "lasers",
            "filter_wheel",
            "stage",
            "zoom",
        ]
```

annie-xd-wang (approved):

PR #542, opened 2023-07-23
Fix Tiger Controller

zacsimile:
Fix issue related to https://asiimaging.com/docs/products/serial_commands#commandwhere_w. Namely, the `WHERE` command always returns axes positions in the order of the underlying hardware, regardless of the order axes are queried. Querying a single axis at a time fixes this issue.

Lots of code cleanup in the process of getting to this answer.

### DIFF
--- a/src/aslm/model/devices/stages/stage_asi.py
+++ b/src/aslm/model/devices/stages/stage_asi.py
@@ -47,7 +47,7 @@ p = __name__.split(".")[1]
 logger = logging.getLogger(p)
 
 
-def build_ASI_Stage_connection(com_port, baud_rate=115200, timeout=1000):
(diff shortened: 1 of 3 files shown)


src/aslm/model/devices/stages/stage_asi.py
```diff
@@ -66,20 +64,10 @@ def build_ASI_Stage_connection(com_port, baud_rate=115200, timeout=1000):
     """
 
     # wait until ASI device is ready
-    block_flag = True
-    wait_start = time.time()
-    timeout_s = timeout / 1000
-    while block_flag:
```
AdvancedImagingUTSW, line 72: Presumably redundant given that we have the auto_redial function in device_startup_functions

AdvancedImagingUTSW (approved):
Looks good.

PR #550, opened 2023-07-28
using asi bu command to get default axes sequence when connect to the ASI stage

annie-xd-wang:


zacsimile (approved):

zacsimile:
Confirmed working on the mesoSPIM. Faster.

PR #552, opened 2023-07-28
Fix tiling

zacsimile:
Also fixes FOV calculation and multiposition append.

Also speeds up the ASI Tiger stage.


src/aslm/controller/sub_controllers/camera_setting_controller.py
```diff
@@ -374,49 +374,61 @@ def calculate_physical_dimensions(self):
...
+
+        #     multi_immersion_focal_length = extended_focal_length / refractive_index
+        #     magnification = tube_lens_focal_length / multi_immersion_focal_length
+        # else:
+        #     magnification = self.parent_controller.configuration["experiment"][
+        #         "MicroscopeState"
+        #     ]["zoom"]
+        #     magnification = float(magnification[:-1])
+
+        # pixel_size = self.default_pixel_size
```
AdvancedImagingUTSW, line 410: Might as well remove the commented code if we have no intention of using it.

src/aslm/model/devices/stages/stage_asi.py
```diff
@@ -166,28 +165,66 @@ def __init__(self, microscope_name, device_connection, configuration, device_id=
...
+
+        if device_connection is not None:
+            ### Speed optimizations
+            # Set speed to 90% of maximum on each axis
+            self.set_speed(percent=0.9)
+
+            # Set feedback alignment
+            # TODO: Grab this dictionary from configuration.yaml as it is per-stage
+            #       In the absence of such a dictionary, set all to 85 by default.
+            feedback_alignment = {"X": 91, "Y": 91, "Z": 91, "T": 85, "V": 88, "W": 88}
```
AdvancedImagingUTSW, line 198: Should this be something we pull from the configuration file in a manner that is analogous to `axes_mapping`?

src/aslm/model/devices/stages/stage_asi.py
```diff
@@ -280,6 +317,16 @@ def move_axis_absolute(self, axis, abs_pos, wait_until_done=False):
             logger.exception(e)
             return False
 
+    def verify_move(self, move_dictionary):
```
AdvancedImagingUTSW, line 320: It seems like we should be doing this regardless, for every stage that we use. We have been burned by this with ASI, Sutter, and PI stages. Should this be evaluated at the `model` or `microscope` level? What do you think @annie-xd-wang?

src/aslm/model/model.py
```diff
@@ -697,7 +697,7 @@ def run_data_process(self, num_of_frames=0, data_func=None):
             Function to run on the acquired data.
         """
 
-        wait_num = 10  # this will let this thread wait 10 * 500 ms before it ends
+        wait_num = 20  # this will let this thread wait 10 * 500 ms before it ends
```
AdvancedImagingUTSW, line 700: Having this thing defined in 4 separate locations seems not super clean. Should we make it an attribute of the model, so that it only defined within __init__?

test/controller/sub_controllers/test_camera_setting_controller.py
```diff
@@ -471,51 +471,30 @@ def test_set_mode(self, mode, readout):
...
         self.camera_settings.parent_controller.configuration["experiment"][
             "MicroscopeState"
         ]["zoom"] = zoom
-        if zoom == "N/A":
-            self.camera_settings.solvent = solvent
 
         # Calling
         self.camera_settings.calculate_physical_dimensions()
 
-        # Checking
```
AdvancedImagingUTSW, line 492: More extraneous code...

AdvancedImagingUTSW (commented):
Given that this works on more than one microscope, I feel pretty good about moving forward. Everything I read makes sense from an abstract perspective. Few line comments that might be worth discussing.

AdvancedImagingUTSW (approved):

Issue #553, opened 2023-07-28
galvo_base - variable not defined


AdvancedImagingUTSW:
While working up a mock-based unit test for the test_galvo_ni device, I realized that `exposure_time` is not defined in the `adjust` method of `galvo_base`...

```
                # galvo Parameters
                try:
                    galvo_amplitude = float(galvo_parameters.get("amplitude", 0))
                    galvo_offset = float(galvo_parameters.get("offset", 0))
                    galvo_frequency = (
                        float(galvo_parameters.get("frequency", 0)) / exposure_time
                    )
                except ValueError as e:
                    logger.error(
                        f"{e} waveform constants.yml doesn't have parameter "
                        f"amplitude/offset/frequency for {self.galvo_name}"
                    )
                    return
```

zacsimile:
It is defined in the `adjust()` method on line 171: `exposure_time = exposure_times[channel_key]`.

PR #554, opened 2023-07-28
Tests for the galvo...

AdvancedImagingUTSW:
See error #553 - I am concerned about the fact that exposure time was not defined in the galvo class, despite being called.

It is now added, but I have learned the hard way that sometimes software works best when things like this aren't fixed. I have asked @annie-xd-wang to review this, but @zacsimile I feel like you should be in the loop. Feel free to reject if we are concerned.


zacsimile:
Since `exposure_time` is only used in the `adjust()` function, there was never any need to make it a class variable. It can be locally scoped. Truthfully, we could make `sweep_time` a local instead of a class variable as well.

AdvancedImagingUTSW:
Now I see it. Strange, I was getting an error… I’ll check later. No daycare today, so my day is pretty wrecked

PR #555, opened 2023-07-28
Delete manufacturer documentation from code base

zacsimile:
This represents a collection of files available elsewhere (plus a few stray documents accidentally committed). I believe we do not need to store these with the code base, since they are available from each instrument's manufacturer.


zacsimile:
Changed my mind.

PR #557, opened 2023-07-28
Set galvo voltage to zero upon exit

### LINKED ISSUE #514: Turn the galvos off on program exit
Set all galvo voltages to 0 on program exit.

mehr0096:


src/aslm/controller/sub_controllers/waveform_popup_controller.py
```diff
@@ -199,7 +199,11 @@ def configure_widget_range(self):
         -------
         None
         """
-        if self.resolution == "high" or self.resolution == "Nanoscale" or self.resolution == "CTASLMv1":
```
AdvancedImagingUTSW, line 202: This seems like something that should be set from the `configuration.yml` file in the GUI parameters section at the bottom...

AdvancedImagingUTSW:
I'll have to try this out on a physical system, but by eye, it looks okay. The only thing that I think could be improved upon is the GUI precision settings. Ideally, these should not be hard-coded by microscope name, but rather configured in the `configuration.yaml` file. @zacsimile, @annie-xd-wang, do you agree?

src/aslm/controller/sub_controllers/waveform_popup_controller.py
zacsimile: Yeah this is old. Black made this change. We should update this, but in a different pull request.

zacsimile:
Yes, but that was a `black` change, not supposed to be a part of this commit. Let's make a new issue for this and solve this in a different PR.

Btw, this is tested on Jinlong/dushyant's setup.

zacsimile:
Just confirmed this works on the v2 as well.

AdvancedImagingUTSW (approved):
Given the confirmation that this works by @zacsimile, we can proceed with the merge.

PR #560, opened 2023-08-02
Fix bdv size

zacsimile:


AdvancedImagingUTSW (approved):
Why did the bit refuse to hang out with the byte?

Because he thought the byte was too big and couldn't handle their small talk!

AdvancedImagingUTSW:
Had to resolve a merge conflict...

PR #561, opened 2023-08-02
Let me enter negative numbers when the bounds are abs(floating point) < 1

zacsimile:


AdvancedImagingUTSW (approved):
So long as type int isn't necessary, seems entirely reasonable.

PR #564, opened 2023-08-02
small issues discovered while I was trying to work with a validated s...

3vwylie:
…pinbox

"w" -> "write" in trace_add()

tk.DoubleVar -> tk.DoubleVar()

### DIFF
--- a/src/aslm/view/custom_widgets/validation.py
+++ b/src/aslm/view/custom_widgets/validation.py
@@ -937,16 +937,16 @@ class ValidatedSpinbox(ValidatedMixin, ttk.Spinbox):
         """ Initialize the spinbox """
         self.resolution = str(kwargs.get("increment", "1.0"))  # Number put into spinbox
         self.precision = self._get_precision()
-        self.variable = kwargs.get("textvariable") or tk.DoubleVar
+        self.variable = kwargs.get("textvariable") or tk.DoubleVar()
         self.required = required
 
         # Dynamic range checker
         if min_var:
             self.min_var = min_var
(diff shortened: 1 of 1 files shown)


AdvancedImagingUTSW:
@3vwylie, what did you do with the GUI to throw the error that you found? I would like to test it out...

3vwylie:
> @3vwylie, what did you do with the GUI to throw the error that you found? I would like to test it out...

I was trying validatedspinboxinstance.variable.get without having a textvariable as a *kwarg. In ASLM, all the other validated spinbox instances that I could find were given a textvariable, so the other side of the "or" was never previously reached. Instead of creating a new doublevar, the code seemed previously to set the variable as the tkinter class for doublevar.... This is approximately the code I was using before I made the changes.  I expected myvar and spinJ1.variable to look similar, which is why I print statemented them.

        spinJ1 = validation.ValidatedSpinbox(
            frm_2,
            from_=0,
            to=25,
            increment=1,
            width=6,
            )
        myvar = tk.DoubleVar()
        print(myvar, spinJ1.variable)

3vwylie:
I thought that I put an image as part of my pull request comment but I guess not :(

AdvancedImagingUTSW:
Confirmed. If the text variable is not specified, then the validated spinbox will try to assign it as a doubleVar, as @3vwylie showed. However, because this lacks the (), you won't be able to retrieve the variable with a .get call. Now fixed. Thanks.

AdvancedImagingUTSW (approved):

PR #570, opened 2023-08-04
Add flipx, flipy and flipz flags to configuration.yaml for stages

### LINKED ISSUE #520: Add flipx/flipy/flipz flags to `configuration.yaml` for stages
Assume by default that increasing X, Y, or Z moves the sample closer to the objectives (further into the field of view). Use this information when saving positions of tiles during acquisition so that the tiled image comes out in the right order. Currently, this is done for `VolumeSearch` and can be changed on a per-scope basis using the plugin architecture for features. We should generalize this behavior.

annie-xd-wang:


AdvancedImagingUTSW:
Confirmed that the stage flip flags now reverse the movement of the stage in X, Y, and Z. Can now collect data backwards! 
image

AdvancedImagingUTSW (approved):
Everything thus far appears to work nicely. Only thing not explicitly tested is the tiling wizard. Presumably should not matter if the flip flags are all False.

PR #571, opened 2023-08-04
Update acquire_bar_controller.py

JinlongL:
Number of channels was incorrect. Now it only counts the channels that are selected.


AdvancedImagingUTSW:
Why was this closed?

PR #581, opened 2023-08-08
Fix stage hot key with flip flags

annie-xd-wang:
Fixed the bug: stage hotkeys were triggered twice.
Fixed the bug: stage hotkeys didn't work with the stage limits flag.
Fixed the bug: stage hotkeys didn't work with the flip flags.


AdvancedImagingUTSW (approved):
Works, and looks good.

PR #583, opened 2023-08-08
Toggle Save Acquisition - Keystroke

AdvancedImagingUTSW:
Keystroke for toggling save now works.


AdvancedImagingUTSW:
Quick thought. This should be in the menu controller

annie-xd-wang (approved):

PR #586, opened 2023-08-08
584 readout direction should default to last used

### LINKED ISSUE #584: Readout Direction Should Default to Last Used
Either the bottom to top, or the top to bottom. 

Always defaults to top to bottom, which can be wrong depending on the microscope.

AdvancedImagingUTSW:


AdvancedImagingUTSW:
Because we both made essentially the same changes (with the exception of the test), merging.

PR #589, opened 2023-08-08
updated constant velocity acquisition to be compatible with new tiger controller functions

mehr0096:
made constant velocity acquisition compatible with new tiger controller functions. Updated velocity calculations


AdvancedImagingUTSW (approved):
Some parameters should probably be pulled from the configuration file, but ok for now

PR #590, opened 2023-08-08
apply camera flip_x and flip_y

### LINKED ISSUE #521: Add flipx/flipy flags to `configuration.yaml` for camera, act accordingly
The camera may be flipped w.r.t. the normal stage orientations (see #520). It may be the case (as in the ctASLMv2) that we need to not only track positions but flip the pixel orientation (the ctASLMv2 camera is upside-down and needs to be saved to disk in reverse order, `img[:,::-1]`).

annie-xd-wang:


AdvancedImagingUTSW (approved):
The flip flags work as anticipated for the real-time display and saving of the data. I'm not entirely convinced that the stage bounds are working properly. When I right-click the camera display, and ask it to move to that spot, it sometimes triggers the stage limits when it shouldn't. Will evaluate further.

PR #608, opened 2023-08-23
Turn off lasers after each image acquisition...

### LINKED ISSUE #604: Turn off lasers whenever data thread is paused
If the data thread is paused, the signal thread can continue moving stages but must turn off the laser until the data thread is resumed. This will prevent our samples from bleaching during long pauses.

AdvancedImagingUTSW:
I think should should be fast enough that we can make the call each image...


zacsimile:
One thing we could consider is performing digital blanking only if there is a digital port. Setting the analog voltages each time we turn the laser on or off can be a bit slower than digital blanking.

zacsimile (approved):

PR #618, opened 2023-09-03
Fix focus out problem.

### LINKED ISSUE #559: Empty tkinter widget throws error for number of pixels
Tried to adjust the size of the camera to make the number of pixels in x == y, just to avoid any problems with saving. Appears that we are not using a validated widget for this widget,  and an empty entry will throw an error:

```2023-08-01 13:40:35,891 - controller - ERROR - camera_setting_controller: could not convert string to float: '' similar to TclError```

No error thrown in the logs. 

[model_debug.log](https://github.com/TheDeanLab/ASLM/files/12232995/model_debug.log)
image

AdvancedImagingUTSW:
I think it was the same problem we faced with the other error. When the user had an empty entry, and focused out, it would consider it valid when it was not. Not at least have a mechanism for handling empty strings in the camera settings controller. Also change the tk GUI features so that they are the validated spinbox that we commonly use.


zacsimile:
Partially fixed, but still an issue on line 225 of `camera_setting_controller.py`. Should we just it so on focus out any box with an empty string reverts to its previous valid value?

zacsimile:
Also now fixes #621. Addresses problem of setting ROIs outside of allowed sizes. E.g. Hamamatsu Orca Flash 4.0 V3 requires ROIs be changed in steps of 4 under certain acquisition modes.

Remaining: unit tests for the new stuff.

image

zacsimile:
I should explain why it went this way. If we just ignore values when this box is left empty, the program crashes on acquisition (it doesn't know the camera ROI size we want). So, instead, it makes sense to return a default value in the empty case. The default value I proposed is the smallest size an ROI can be. This is dependent on the camera type and its mode. It was cleaner to implement better ROI alignment than to just set the minimum to the Hamamatsu Orca Flash 4.0 V3 minimum. Dynamic choice of ROI sizes is now implemented for Hamamatsu cameras and coded into the base class. It still needs to be implemented for other cameras.

zacsimile:
This works. Tested on multiple system. We still need to mock the `HamamtsuAPI` object for testing in order to fix the rounding behavior in place. That will be a different PR.

zacsimile (approved):

PR #631, opened 2023-09-28
End acquisition on stop call

### LINKED ISSUE #626: Laser doesn't always turn off on stop acquisition if multiple channels are selected
For example, in continuous mode, if multiple channels are selected, one of the lasers may stay parked on a sample after the acquisition is stopped.

zacsimile:
Closes #626.


AdvancedImagingUTSW (approved):
Interesting. If it works on the microscopes, and passes the tests, we can merge.

PR #634, opened 2023-09-29
a few autofocus tests

AdvancedImagingUTSW:
Some basic examples on how to make a test...

The test directory mirrors the src directory.  For autofocus.py, we thus have a file called test_autofocus.py. 

I used the unittest package to build the tests, but pytest is easier and has more functionality built in (for example, by allowing you to easily run tests in loops with parameterize, etc.). 

I did not go into testing things like pre_func here since that is more involved. 

So long as these tests pass our continuous integration pipeline, you can review the changes, approve them or recommend changes, and we can merge it.


AdvancedImagingUTSW:
@conorhughmcfadden - does this make sense?

conorhughmcfadden (approved):
Yep. Looks like some autofocus tests!

PR #639, opened 2023-10-13
Update waveform_popup_controller.py

### LINKED ISSUE #638: Off by one error in number of pixels for galvo "Estimate Frequency"
The "Estimate Frequency" button appears to estimate the perfect frequency for Number of Pixels - 1.

E.g., estimated for 159...
image

...but works for 158
image

AdvancedImagingUTSW:
See if this fixes the `estimate frequency` function...

### DIFF
--- a/src/aslm/controller/sub_controllers/waveform_popup_controller.py
+++ b/src/aslm/controller/sub_controllers/waveform_popup_controller.py
@@ -548,6 +548,10 @@ class WaveformPopupController(GUIController):
             # If we are not in the light-sheet mode, widget returns an empty string.
             return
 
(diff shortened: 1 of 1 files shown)


zacsimile:
This should fix it. Is it the most mathematically correct solution? I wonder if we should instead remove the `-1` in `get_camera_line_interval_and_exposure_time()`, though I think that is correct for a convolution.

AdvancedImagingUTSW:
I think it is more mathematically correct without the `+1`, but I would have to think about the `-1` in the `get_camera_line_interval_and_exposure_time()`. 

The big picture is to sweep the beam twice during the effective pixel dwell time.

zacsimile (approved):

PR #645, opened 2023-10-24
Fix defocus issue in `ZStackAcqusition`

### LINKED ISSUE #623: Defocus channel option appears to not propagate in z-stack mode
Reason needs to be chased down. This prevents us from taking nice multicolor stacks where both channels are in focus. The workaround is to acquire each channel individually.

zacsimile:
Closes #623.


AdvancedImagingUTSW (changes requested):
You told me to not to merge this.

PR #646, opened 2023-10-25
Add defocus to `ZStackAcquisition`

### LINKED ISSUE #623: Defocus channel option appears to not propagate in z-stack mode

zacsimile:
Closes #623.


zacsimile:
Works on the mesoSPIM BT.

AdvancedImagingUTSW (approved):

PR #655, opened 2023-10-27
651 add shearing transform to bdv

### LINKED ISSUE #651: Add Shearing Transform to BDV
If you look at [https://imagej.net/plugins/bigstitcher/#2d-multi-tile-dataset-28-mb](https://imagej.net/plugins/bigstitcher/#2d-multi-tile-dataset-28-mb), the example data sets all have split affine transform matrices, which stack on top of one another. So we need to add like `Translation to Regular Grid` to our current translation matrix. And then we could have `Shear` as an optional extra transform that we inject in the case a shear is defined with our scope, maybe from configuration.yaml

```
<ViewRegistration timepoint="0" setup="8">
      <ViewTransform type="affine">
        <Name>Tile ICP Refinement</Name>
        <affine>1.0009775557493341 1.6889894237572095E-4 4.433822850283759E-5 0.8704882273809396 0.004065029174062048 0.9990351787796182 9.854905708783271E-4 -1.1427728541768902 0.0010688510520953645 -0.0013111569993652504 1.0008616832683797 -0.4831841089543653</affine>
      </ViewTransform>
      <ViewTransform type="affine">
        <Name>Stitching Transform</Name>
        <affine>1.0 0.0 0.0 0.675940480301179 0.0 1.0 0.0 1.3719299181531284 0.0 0.0 1.0 -0.13486861011732687</affine>
      </ViewTransform>
      <ViewTransform type="affine">
        <Name>Translation to Regular Grid</Name>
        <affine>1.0 0.0 0.0 -486.4 0.0 1.0 0.0 -255.99999999999994 0.0 0.0 1.0 -191.11111111111111</affine>
      </ViewTransform>
      <ViewTransform type="affine">
        <Name>calibration</Name>
        <affine>1.0 0.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 0.0 4.444444444444445 0.0</affine>
      </ViewTransform>
    </ViewRegistration>
```

AdvancedImagingUTSW:
Shear confirmed to work. Need to test rotation on the microscope. 

Posting as a draft should you want to provide any feedback.


zacsimile:
Seems more or less ok. Why can we only rotate in one dimension at at time? Why do we add a shear/rotation affine transform even if there is no rotation specified?

AdvancedImagingUTSW:
For the time being, I could only think of applications where we would shear or rotate one axis at a time. Figured we could always add more dimensions should we need to in the future.

As for including additional transforms in the xml, I didn't think it would matter so I avoided conditional statements. Now included in the latest push.

zacsimile:
it seems like more work to not implement arbitrary rotation in all three dimensions here. That would just be a product of all the matrices.

You could save a lot of code (and computational time) by writing `cost = np.cos(np.deg2rad(self.rotate_angle))` and `sint = np.sin(np.deg2rad(self.rotate_angle))` at the top of the function and using these instead of re-calculating the sine/cosine of the angle multiple times. Since the shear/rotation matrices are calculated only once, you could move the raveling inside their functions and store the matrices as flat. I'd recommend the function signature be `calculate_rotate_transform(self, xdeg=None, ydeg=None, zdeg=None)` and only calculate the rotation about each axis if `{axis}_deg` is not None. Then you `np.matmul` (`x @ y`) the rotation matrices that are not `None` together to create the rotation matrix.

AdvancedImagingUTSW:
Pulled out the cosine and sine calculations, although they were on the microsecond timescale. 

I did not implement the multi-directional rotation for more than one direction. Although the abstraction is nice, it is more than we need at the moment.

AdvancedImagingUTSW:
I'll fix the tests when I get a chance to test it.

zacsimile (approved):
This works! We can clean it up a bit later. Fine to merge as is.

Notes for later:

* We don't need `shear_data` and `shear_angle` as `shear_data =shear_angle != 0`.
* We don't need to compute the `self.shear_transform` or `self.rotate_transform` matrices as `np.eye(3,4)` in the event we are not adding them to the metadata.
* The `affine_rotation` function could be written a bit shorter.

PR #658, opened 2023-10-31
Introduce new tiling wizard and controller

### LINKED ISSUE #628: Tiling wizard gridding appears to be shifted by half a frame

zacsimile:
This is designed to allow for tiling along X, Y, Z and F. The goal is to image a full tissue on the ctASLM v1/v2, where there is both a z-stacking stage (Z) and a sample scanning stage in the z-direction (F).

This works, but in order for BDV to display this as a contiguous tile, we will need to add Z and F together in the affine transform matrix for just these two microscopes. 

Also closes #628.


AdvancedImagingUTSW (approved):
Looks good to me. I adjusted the numpydoc to the format I found sphinx works best with. Pretty minor detail...  Nice work getting this all dialed in.

PR #663, opened 2023-11-06
657 cleanup requirementstxt andor replace with pyprojecttoml

### LINKED ISSUE #657: Cleanup Requirements.txt and/or replace with pyproject.toml
I'm pretty sure we no longer need tensor flow in the requirements, which would clean up the dependency install substantially. Could do this while updating to a pyproject.toml, if we think that is a good idea.

Also, the Capstone team will need this library: https://github.com/Mecademic/mecademicpy Not sure if we should add this to our requirements, add it as a sub-module, or install a static version in our API folder. Open for opinions.

AdvancedImagingUTSW:


zacsimile:
I think we need to update `.github/workflows` scripts to use the `pyproject.toml` install to get the tests to pass.

AdvancedImagingUTSW:
Now passing @zacsimile. Removed all GPU calls since we removed the Copy and Tensorflow packages.

zacsimile (approved):

PR #667, opened 2023-11-09
Verify that multiposition is actually true

### LINKED ISSUE #665: Acquisition jams if multiposition is selected but multiposition table is empty
Need to write a check for this and alert the user that multiposition is selected, but the table is empty.

AdvancedImagingUTSW:
Prior to launching the feature container with the model, we now verify that the multi position table actually has values if the multi position toggle is selected. If not, we disable the toggle in the GUI and set the multi position to False.

If this is decent, will have to think about how to put a test in...

### DIFF
--- a/src/aslm/controller/controller.py
+++ b/src/aslm/controller/controller.py
@@ -1076,6 +1076,10 @@ class Controller:
                 self.channels_tab_controller.is_multiposition_val.set(True)
                 self.channels_tab_controller.toggle_multiposition()
 
+            elif event == "disable_multiposition":
+                self.channels_tab_controller.is_multiposition_val.set(False)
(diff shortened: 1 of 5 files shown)


AdvancedImagingUTSW:
This would have been easier if I could patch the `model` with a `MagicMock` version the `multiprocessing.event_queue`, but I would have had to patch the local instance of `model` which is a pytest fixture and things got weird. 

Instead, I created a whole new instance of the model, which was a bit annoying, and this time explicitly made the mock event_queue. I then test how the mock is called when the multi position table is populated.

zacsimile:
Yeah I tried the other way you mentioned and could not get it to work either. Thanks for coming up with this solution.

PR #669, opened 2023-11-15
Add stop stage to tiling wizard controller

### LINKED ISSUE #518: Force position call when setting tiling wizard bounds, z-stack bounds
Whenever we set the z-stack bounds or the bounds of the the tiling wizard, we should force a call to the stage for that axis position.

zacsimile:
Closes #518.


zacsimile:
Confirmed working on the v2.

AdvancedImagingUTSW (approved):

PR #678, opened 2023-11-16
Deal with buffer overrun in HamamatsuAPI

zacsimile:
This fixes the issue in #674 for the Hamamatsu cameras. This will have to be fixed for Photometrics, etc., independently.


zacsimile:
Further investigation reveals this is not at present an issue with Photometrics.

AdvancedImagingUTSW (approved):

PR #679, opened 2023-11-16
Updated hardware

AdvancedImagingUTSW:
Just docs.

### DIFF
--- a/src/aslm/model/devices/daq/daq_base.py
+++ b/src/aslm/model/devices/daq/daq_base.py
@@ -73,10 +73,10 @@ class DAQBase:
 
         # Initialize Variables
         #: float: Sample rate of the DAQ
-        self.sample_rate = self.daq_parameters["sample_rate"]
+        self.sample_rate = None
 
         #: float: Sweep time of the DAQ
-        self.sweep_time = self.daq_parameters["sweep_time"]
+        self.sweep_time = None
 
         # Remote Focus Parameters
         #: dict: Dictionary of remote focus ramp falling percentages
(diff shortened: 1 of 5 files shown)


src/aslm/model/devices/daq/daq_base.py
```diff
@@ -73,10 +73,10 @@ def __init__(self, configuration):
 
         # Initialize Variables
         #: float: Sample rate of the DAQ
-        self.sample_rate = self.daq_parameters["sample_rate"]
+        self.sample_rate = self.daq_parameters.get("sample_rate", 100000)
 
         #: float: Sweep time of the DAQ
-        self.sweep_time = self.daq_parameters["sweep_time"]
+        self.sweep_time = self.daq_parameters.get("sweep_time", 0.2)
```
zacsimile: I think the only place this gets used is in `daq_ni.create_camera_task()`. Every other place re-defines `sweep_time` before calling it. If we get into a situation where this is not re-defined prior to calling and the exposure time is longer than the sweep time, this may create issues. Replacing this with `self.sweep_time = None` would at least alert us to this issue, should it arise. Leaving this with a default of `0.2` would possibly create silent failures, with the microscope behaving improperly with no warning.

AdvancedImagingUTSW:
I can remove those changes. I was just hoping to eliminate some unnecessary configuration entries at the time.

zacsimile:
I would remove them all the way and set them to `None` in `__init__`.

AdvancedImagingUTSW:
I abandoned the change so that this PR is now only documentation.

zacsimile (approved):

PR #688, opened 2023-11-20
637 dichroic control

### LINKED ISSUE #637: Dichroic Control
Need this for the spectral TIRF.

AdvancedImagingUTSW:
This is a partially complete PR. I will first review the code myself on here. 

It now allows the number of channels to be defined in the configuration file, as we wanted but never properly implemented. A test for this functioning should be accompanied with this PR by me. 

If a dichroic turret is included in the PR, an additional dropdown is now added.

I need to vet the operation of the dichroic changer, but it seemed to work when I tested it.

It is incomplete, since I do not add the option to interface with the dichroic changer in the model. Only sub-controllers, etc.


AdvancedImagingUTSW:
For some reason, these tests are now stalling.

AdvancedImagingUTSW:
The tests pass. I don't know why pushing the coverage report to codecov hangs on this branch only. Super confused.

AdvancedImagingUTSW:
Essentially, @zacsimile and I decided that this is probably not implemented as we would like. As the first of its type, where the GUI responds dynamically to the configuration, we have to be careful and implement the architecture as we want so that we can build more effectively from this foundation. Currently, by passing the configuration to the view, we create multiple references to objects, and these references are circular. Therefore, on garbage collection, elimination of object references is non-ideal. This could get worse if we build upon this principle. Logic should be moved to the controller/sub-controllers.

Essentially, populate everything in the VIEW, and selectively disable them in the controller according to configuration file for that microscope.

PR #696, opened 2023-11-29
592 ctaslmv1camera starts acquiring data before stage moves to starting position

### LINKED ISSUE #592: (ctASLMv1)Camera starts acquiring data before stage moves to starting position
While doing Z stack acquisition, user moves stage to the starting position, mark it as start pos, then move to the end position, and mark it as end pos. Without bringing the z piezo back to the starting position before imaging, end position will be the first frame. In single channel image, we can bring the z piezo back to the starting position before the acquisition, so the start position will be the first frame. However, when doing multi-channel imaging, bring the z piezo back to the start position before starting the whole acquisition will only have the start position in the first frame in the first channel, other channels will still have the end position in the first frame. The frame number or the slice number will be the same for all channels, resulting the n frame of the first channel be the n+1 frame for other channels. 
Capture

annie-xd-wang:
Please test on the machine before merging it.

### DIFF
--- a/src/aslm/model/devices/daq/daq_base.py
+++ b/src/aslm/model/devices/daq/daq_base.py
@@ -75,8 +75,11 @@ class DAQBase:
         #: float: Sample rate of the DAQ
         self.sample_rate = self.daq_parameters["sample_rate"]
 
-        #: float: Sweep time of the DAQ
-        self.sweep_time = self.daq_parameters["sweep_time"]
+        #: dict: Sweep times for different channels
+        self.sweep_times = None
+        
+        #: dict: exposure times for different channels
+        self.exposure_times = None
 
         # Remote Focus Parameters
         #: dict: Dictionary of remote focus ramp falling percentages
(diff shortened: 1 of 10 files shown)


AdvancedImagingUTSW (commented):
Looked over the code additions and they seem reasonable. Will move to testing them immediately.

AdvancedImagingUTSW:
As far as I can tell, this works on CT-ASLM-V2. I don't think it should interfere with operation on any other microscope. @zacsimile, would you like secondary testing on another system?

zacsimile:
If it works, I'm fine with merging it. My one concern is this wasn't a problem specific to the piezo stage, but rather a problem that just manifests in the piezo stage. It seems we generate an extra batch of waveforms at the start of every z-stack. The fix here seems to address this only for the piezo stage. If that's sufficient, I'm fine with merging it.

annie-xd-wang:
It is correct to generate a batch of waveform before every z-stack. This DAQ analog task before every z-stack reruns in the whole stack. The problem here is that we rely on a DAQ analog task to move the piezo stage, and there will be a DAQ task writing every time we move the stage. So the waveform we generate before a stack runs once and at that time it doesn't set the correct voltage for the piezo stage. The current ZStackAcquisition scenario should work for other kinds of stages that are not driven by a DAQ analog task.

AdvancedImagingUTSW (approved):

PR #697, opened 2023-11-29
Update stage_galvo.py

AdvancedImagingUTSW:
@annie-xd-wang - Now the stage settle duration is fully functional. Could you take a look at this and make sure it is not too awful? Happy to jump on a zoom call and discuss...


AdvancedImagingUTSW:
Annie wants to know the following:

Hi Kevin, just updated the code. Please test it. The three most important tests are 1, start the software and see whether there is an initial voltage sent from the DAQ to the stage. 2. Run in "continuous" mode and move the stage to see whether the voltage could change( the stage should move). 3. Change between modes "confocal-projection" and "continuous" to see whether the software works correctly.

Please let me know if there is any error, especially whether the DAQ supports two analog tasks on a single board.

AdvancedImagingUTSW:
While working with Bo-Jui to find a source of latency between adjacent image acquisitions, we focused in on the `stage_galvo.py` file. 

To move the stage as part of a common_feature takes ~26 ms. Of which, approximately 20 ms of this due to the `move_axis_absolute` method in `stage_galvo.py`. 

We did some timing tests to figure out the source. A couple shared dictionary calls show up...

3 ms. 
```
        microscope_state = self.configuration["experiment"]["MicroscopeState"]
```

3 ms.
```
            channel = microscope_state["channels"][channel_key]
```

The first block of code after if channel["is_selected"] is True, including the Try/Except statement, also costs 2 ms. 

17 ms
```
        self.daq.analog_outputs[self.axes_channels[0]] = {
            "trigger_source": self.trigger_source,
            "waveform": self.waveform_dict,
        }
```

3 ms
```
        self.daq.update_analog_task(self.axes_channels[0].split("/")[0])
```

Long story short, I think once we get the PR finalized and functioning, we will probably also simultaneously improve the speed of the software.

AdvancedImagingUTSW:
@annie-xd-wang - I created a branch from this, updated from develop, handled the conflicts, then updated this branch from that branch.

While running in a continuous mode, the stage worked properly.

annie-xd-wang:
We need to test acquisitions in both ZStack and confocal-projection modes.

AdvancedImagingUTSW:
The z-stack mode works. The projection mode does not:

```
No key 'GalvoNIStage' object has no attribute 'x_min' in move_dictionary or axis missing from ['z'].
No key 'GalvoNIStage' object has no attribute 'x_min' in move_dictionary or axis missing from ['z'].
No key 'GalvoNIStage' object has no attribute 'x_min' in move_dictionary or axis missing from ['z'].
exposure_time Configuration Failed 0.00247066 0.0025941315789473683
```

annie-xd-wang (approved):

PR #699, opened 2023-12-01
Adaptive Optics final merge (at the buzzer)

conorhughmcfadden:
Merging all of the Adaptive Optics code. If there is a mirror defined in the config file, an "Adaptive Optics" option appears in the feature menu, which opens a GUI capable of doing iterative AO.

Main AO routine in adaptive_optics.py is very similar to the routine found in autofocus.py. Main API is in imagineoptics/imop.py, custom built from ctypes using the manufacturer DLL.

Mirror is a Mirao52e from ImagineOptics, using the WaveKit API.


AdvancedImagingUTSW:
The only major thing that pops out to me is that the logic for the mirror is in the controllers, not the model. This isn't how we have traditionally done it. Typically all of the logic and hardware interactions are done in the model. The controller just communicates between the view and the model. Also, some of the basic commands in the adaptive optics feature could be placed in the tools section.

I didn't completely finish documenting the feature, had to jump on the flight.

zacsimile:
Confirmed working on the mesoSPIM BT.

annie-xd-wang (approved):
The code looks fine to me. I changed two things: 1) It won't load a synthetic mirror if there is no mirror in the configuration yaml file. 2) The code to change the camera. The two changes will not break adaptive optics, but running it on a machine to double-check is better.

AdvancedImagingUTSW (approved):

PR #707, opened 2023-12-05
Update the icons.

AdvancedImagingUTSW:
I need to look into the mic.icon image. That might be Windows-specific.


AdvancedImagingUTSW:
For some reason, the relative path that we previously used does not work on my computer anymore. Consequently, it throws the except statement in the `SplashScreen` 
image

AdvancedImagingUTSW:
I confirmed that the new path works on a Windows machine, but it didn't work on a Mac...

Fixed now. Confirmed on both Mac and Windows machines.
image

zacsimile (approved):

PR #708, opened 2023-12-06
queue all calls to the model

### LINKED ISSUE #677: Model called twice
It's time to add a `_Custody` object to prevent these kind of errors.

```
model thread exception happened! Two different threads tried to use the same ObjectInSubprocess at the same time! This is bad. Look at the docstring of concurrency_tools.py to see an example of how to use a _Custody object to avoid this problem. Traceback (most recent call last):
  File "c:\users\dean-lab\desktop\aslm\src\aslm\controller\thread_pool.py", line 251, in func
    target(*args, **kwargs)
  File "c:\users\dean-lab\desktop\aslm\src\aslm\controller\controller.py", line 641, in <lambda>
    "model", lambda: self.model.run_command("update_setting", *args)
  File "c:\users\dean-lab\desktop\aslm\src\aslm\model\concurrency\concurrency_tools.py", line 404, in __getattr__
    with self._.parent_pipe_lock:
  File "c:\users\dean-lab\desktop\aslm\src\aslm\model\concurrency\concurrency_tools.py", line 578, in __enter__
    raise RuntimeError(
RuntimeError: Two different threads tried to use the same ObjectInSubprocess at the same time! This is bad. Look at the docstring of concurrency_tools.py to see an example of how to use a _Custody object to avoid this problem.
```

annie-xd-wang:
#677

### DIFF
--- a/src/navigate/controller/controller.py
+++ b/src/navigate/controller/controller.py
@@ -660,30 +660,6 @@ class Controller:
                 "model", lambda: self.model.run_command("stage_limits", *args)
             )
 
-        # mirror commands:
-        elif command == "flatten_mirror":
-            self.model.run_command("flatten_mirror", *args)
-        elif command == "zero_mirror":
-            self.model.run_command("zero_mirror", *args)
-        elif command == "set_mirror":
-            self.model.run_command("set_mirror", *args)
-        elif command == "set_mirror_from_wcs":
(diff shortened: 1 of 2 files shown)


zacsimile:
Thanks, Annie! Looks right and seems to work on the scopes :).

zacsimile (approved):

PR #726, opened 2023-12-12
680 accurate imaging time estimation

### LINKED ISSUE #680: Accurate imaging time estimation
Every user asks me how long an acquisition is going to take. Due to inconsistencies in stage response times (e.g. on the ASI) our current estimation of an experiment's duration is fairly inaccurate. I propose we add a file to `.ASLM` that keeps track of the true acquisition time and conditions for the last `N` (10? 20?) acquistions and uses this information to provide more accurate estimates of the duration.

AdvancedImagingUTSW:


zacsimile:
Tested this on the mesoSPIM BT. It seems to work. I think we should get rid of the other time estimate as it seems quite off by comparison.

AdvancedImagingUTSW:
I plan on looking at how that estimate is calculated to see if we can improve it. If it isn't easy, then I'll go ahead and rip those widgets out and any of the logic that is accompanied by it.

AdvancedImagingUTSW:
Going to merge what changes we have, but will continue to try to implement a better timing estimate. Low hanging fruit in the estimate would include using the cycle time, not the exposure time, to calculate the duration of imaging. This still does not account for any overhead that result from serial communication or stage movement, but it will be a step in the right direction. 

To move further in the right direction, we could either take the approach that @zacsimile recommended, or make sure that every stage has a stage velocity attribute, which we could then use to estimate the duration for a move depending on the distance of that move.  Still doesn't account for serial communication though.

The last approach that I can think of is to have a wait ticket for each stage move, similar to the camera exposure. If say there is an attribute in each stage class that reports the average duration of time to make a stage move, which would account for the serial communication, we could use that average...

@zacsimile, since you already looked at this and tested it, you want to approve it? Once it is merged we can reopen the issue that is linked with it.

zacsimile (approved):

zacsimile:
I don't think we need to add any additional wait tickets. The current one should take stage movement into account. I think what you have done in this PR should work reasonably well on any scope, and we can just extend this logic to the other time estimation panel. In the event an accurate frame rate is not available, we can use cycle time instead.

PR #769, opened 2023-12-29
227 automatic handling of synthetic stage axes gui should only display available axes

### LINKED ISSUE #227: Automatic Handling of Synthetic Stage Axes, GUI should only display available Axes

AdvancedImagingUTSW:
I want to see how this looks on a Windows machine.

### DIFF
--- a/src/navigate/config/configuration.yaml
+++ b/src/navigate/config/configuration.yaml
@@ -121,7 +121,7 @@ microscopes:
       hardware:
         -
           name: stage
-          type: PI
+          type: syntheticstage
           serial_number: 119060508
           axes: [x, y, z, theta, f]
           axes_mapping: [1, 2, 3, 4, 5]
@@ -355,7 +355,7 @@ microscopes:
           min: None
         -
           name: stage
-          type: Thorlabs
+          type: synthetic
           serial_number: 74000375
           axes: [f]
           axes_mapping: [1]
--- a/src/navigate/config/configuration.yaml
+++ b/src/navigate/config/configuration.yaml
@@ -121,7 +121,7 @@ microscopes:
       hardware:
         -
           name: stage
-          type: syntheticstage
+          type: PI
           serial_number: 119060508
           axes: [x, y, z, theta, f]
           axes_mapping: [1, 2, 3, 4, 5]
@@ -355,7 +355,7 @@ microscopes:
           min: None
         -
           name: stage
-          type: synthetic
+          type: Thorlabs
           serial_number: 74000375
           axes: [f]
           axes_mapping: [1]
--- a/src/navigate/controller/sub_controllers/stage_controller.py
+++ b/src/navigate/controller/sub_controllers/stage_controller.py
@@ -64,7 +64,7 @@ class StageController(GUIController):
              The main view of the microscope
          canvas : tkinter.Canvas
              The canvas of the microscope
-         parent_controller : navigate.controller.microscope_controller.MicroscopeController
+         parent_controller : navigate.controller.Controller
              The parent controller of the stage controller
         """
         super().__init__(view, parent_controller)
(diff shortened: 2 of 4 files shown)


AdvancedImagingUTSW:
@zacsimile - Want your opinion. Do you think it would be better to destroy the axes if they are not available, or disable them?

AdvancedImagingUTSW:
This is what it looks like on a Windows machine... If everyone likes how this looks, I can proceed and fix the tests, and also make sure that the spinbox is also disabled. I prefer this over deleting the buttons, since things start to look pretty odd when they go missing (also requires more code to repopulate it if we switch to non-synthetic stages).
image

src/navigate/model/microscope.py
```diff
@@ -276,23 +276,24 @@ def __init__(
             if device_ref_name not in devices_dict["stages"]:
                 logger.debug("stage has not been loaded!")
                 raise Exception(
-                    "Stage not found. \n"
+                    f"Stage not found. "
```
zacsimile: Why the f string?

src/navigate/controller/sub_controllers/stage_controller.py
```diff
@@ -167,6 +172,8 @@ def stage_key_press(self, event):
     def initialize(self):
         """Initialize the Stage limits of steps and positions."""
         config = self.parent_controller.configuration_controller
+        self.disable_synthetic_stages(config)
```
zacsimile, line 175: This probably also needs to be called when switching microscopes.

src/navigate/config/configuration.yaml
```diff
@@ -121,7 +121,7 @@ microscopes:
       hardware:
         -
           name: stage
-          type: PI
+          type: syntheticstage
```
zacsimile: We probably shouldn't update `configuration.yaml` in the main branch.

src/navigate/model/microscope.py
AdvancedImagingUTSW: I think my goal was to provide a more informative Exception and state which stage was not found. I'll go back and do that. Given that this is one of the fragile break points in getting the code set up, I want all feedback to be as useful as possible in diagnosing the problem.

AdvancedImagingUTSW: If you look at the last line of this string, there is a variable for the f-string, but I see it is formatted incorrectly. I will fix it.

src/navigate/config/configuration.yaml
AdvancedImagingUTSW: I accidentally did this while testing it.

src/navigate/controller/sub_controllers/stage_controller.py
AdvancedImagingUTSW: I created a second `microscope` on the mesoSPIM with only synthetic stages. 

The first microscope is still defined with the ASI stages:
image

Here is the one with synthetic stages:
image

And, when I switch back to the original mesoSPIM, all of the stage axes become active again. Nice call @zacsimile

AdvancedImagingUTSW:
I currently think that this is functional. If everyone likes the addition, I will move to add a unit test or two and convert it to a final PR. I just made one more push to make sure that stages can be read in both as a DictProxy and as a ListProxy. Specifically, the difference between this configuraiton file:
```
    stage:
      hardware:
        name: stage
        type: SyntheticStage
        serial_number: 0000
        axes: [x, y, z, f, theta]
        feedback_alignment: [88, 91, 91, 91, 85]
        axes_mapping: [X, V, Z, Y, T]
        volts_per_micron: None
        axes_channels: None
        max: 5
        min: 0
```
And this configuration file:
```
    stage:
      -
        hardware:
          name: stage
          type: SyntheticStage
          serial_number: 0000
          axes: [x, y, z, f, theta]
          feedback_alignment: [88, 91, 91, 91, 85]
          axes_mapping: [X, V, Z, Y, T]
          volts_per_micron: None
          axes_channels: None
          max: 5
          min: 0
```

zacsimile (approved):

PR #772, opened 2023-12-30
Update software_installation.rst

### LINKED ISSUE #771: Minimum recommended specifications
I was thinking that it would be a good idea to list the minimum specifications that we recommend. The most important one is read/write speed with an SSD...

AdvancedImagingUTSW:
This is probably a bit verbose...

### DIFF
--- a/docs/source/software_installation.rst
+++ b/docs/source/software_installation.rst
@@ -24,8 +24,14 @@ Operating System Compatibility
 Hardware Considerations
 -----------------------
 
-.. note::
-   The read/write speed of the operating system drive and data storage is crucial. The **navigate** software leverages pyramidal file formats, which require additional overhead for down-sampling operations. Coupled with the demands of larger sCMOS sensors, this necessitates robust CPU performance and rapid data saving capabilities to ensure efficient processing and storage of large datasets.
+.. important::
+   The read/write speed of the operating system drive and data storage is crucial.
+   The **navigate** software leverages pyramidal file formats, which require additional overhead for down-sampling operations.
+   Coupled with the demands of larger sCMOS sensors, this necessitates robust CPU performance and rapid data saving capabilities
+   to ensure efficient processing and storage of large datasets. For example, for a standard Hamamatsu camera with a 2048 x 2048 sensor,
+   operating at 16-bit depth and 20 frames per second, the data save rate is approximately ~170 MB/s. With additional down-sampling of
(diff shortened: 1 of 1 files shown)


docs/source/software_installation.rst
```diff
@@ -2,6 +2,87 @@
...
+   The software remains untested on Linux systems. Therefore, we cannot guarantee its functionality or performance on Linux platforms at this time. Users considering the use of **navigate** software on Linux should proceed with caution and be prepared for potential compatibility issues, especially with respect to device drivers.
+
+.. note::
+   For optimal performance and compatibility, it is strongly recommended to run the **navigate** software on a Windows machine, adhering to the specified system specifications.
+
+Hardware Considerations
+-----------------------
+
+.. note::
+   The read/write speed of the operating system drive and data storage is crucial. The **navigate** software leverages pyramidal file formats, which require additional overhead for down-sampling operations. Coupled with the demands of larger sCMOS sensors, this necessitates robust CPU performance and rapid data saving capabilities to ensure efficient processing and storage of large datasets.
```
zacsimile: Maybe a slightly more detailed explanation? HDDs write at about 100 MB/s, while SSDs can do 500 MB/s. If we go for 20 fps (50 ms integration time) on a 2048 x 2048 camera with 16-bit depth, we're already around 168 MB/s write speed, which would require SSD. If we add on top of it the 4 down samplings, that's `20*(2048*2048+1024*1024+512*512+256*256)*2*1e-6` ~ 223 MB/s. We need the SSDs even w/o the pyramidal format if we want to go fast.

AdvancedImagingUTSW: Just added in latest push.

docs/source/software_installation.rst
```diff
@@ -2,6 +2,93 @@
...
+
+Hardware Considerations
+-----------------------
+
+.. important::
+   The read/write speed of the operating system drive and data storage is crucial.
+   The **navigate** software leverages pyramidal file formats, which require additional overhead for down-sampling operations.
+   Coupled with the demands of larger sCMOS sensors, this necessitates robust CPU performance and rapid data saving capabilities
+   to ensure efficient processing and storage of large datasets. For example, for a standard Hamamatsu camera with a 2048 x 2048 sensor,
+   operating at 16-bit depth and 20 frames per second, the data save rate is approximately ~170 MB/s. With additional down-sampling of
```
zacsimile: 170 is already outside HDD capabilities. I think making this about the pyramidal data is a bit of red herring.

zacsimile (approved):

PR #776, opened 2024-01-03
764 modify saving approach during custom acquisition

### LINKED ISSUE #764: Modify saving approach during custom acquisition
Ideally checking "Save Data" should work the same in "Customized" mode as in ZStackAcquisition. Currently we use a custom writer in Customized mode.

annie-xd-wang:


AdvancedImagingUTSW:
Tested with @zacsimile using his feature where empty tiles are eliminated and saved on the fly, and it works perfectly for OME, OME-TIFF, N5, and HDF5.

zacsimile (dismissed):

AdvancedImagingUTSW (approved):

PR #779, opened 2024-01-04
Saving in registered acquisition modes

annie-xd-wang:


annie-xd-wang:
The confocal projection plugin is ready to test with this branch. In this branch, I add functions to set saving shape info for acquisition modes from plugins.

AdvancedImagingUTSW:
@zacsimile - @annie-xd-wang and I tested this on CT-ASLM-V2. Plugin requires a galvo-ni stage to be configured as the z-stage, but otherwise works perfectly. Annie will correct the tests, and I think we can merge.

AdvancedImagingUTSW (approved):

PR #797, opened 2024-01-12
Uninstall plugins

annie-xd-wang:


AdvancedImagingUTSW:
Tested with @annie-xd-wang and it worked beautifully. Once we have documentation with a few screenshots on how to install and uninstall a plugin, we can merge.

annie-xd-wang:
Not very confident about where to put the install/uninstall plugin document. It's under 'User Guide->acquiring data`.

PR #802, opened 2024-01-17
800 eliminate projection mode

### LINKED ISSUE #800: Eliminate projection mode
Currently redundant. Remove projection mode, and keep confocal projection mode as a plugin which is fully functional currently.

annie-xd-wang:


AdvancedImagingUTSW (dismissed):
Tested on microscope. Everything works, except for the timer in the acquisition bar when in the customized mode. It shows a bogus number.

Issue #630, opened 2023-09-27
Camera Exposure/Line Interval Error


AdvancedImagingUTSW:
Will follow up with a traceback tomorrow, but this line of code in `microscope.prepare_next_channel()` is throwing an error...

```
# Camera Settings
        self.current_exposure_time = float(channel["camera_exposure_time"])
        if (
            self.configuration["experiment"]["CameraParameters"]["sensor_mode"]
            == "Light-Sheet"
        ):
            (
                self.current_exposure_time,
                camera_line_interval,
            ) = self.camera.calculate_light_sheet_exposure_time(
                self.current_exposure_time,
                int(
                    self.configuration["experiment"]["CameraParameters"][
                        "number_of_pixels"
                    ]
                ),
            )
            self.camera.set_line_interval(camera_line_interval)
        self.camera.set_exposure_time(self.current_exposure_time)
```

mehr0096:
```exposure_time Configuration Failed 0.00743218 0.0075806400000000005
Traceback (most recent call last):
  File "c:\users\dean-lab\documents\aslm\src\aslm\model\devices\daq\daq_ni.py", line 497, in stop_acquisition
    self.camera_trigger_task.stop()
AttributeError: 'NoneType' object has no attribute 'stop'```

zacsimile:
@mehr0096 I think this is a different error, related to the DAQ rather than the camera. The error @AdvancedImagingUTSW mentioned has to do with a mismatch in the set exposure time/line interval.

AdvancedImagingUTSW:
I was the one who posted that wrong traceback previously. Not sure what I was thinking at the time.

Looking into this issue.  Presumably, this issue is happening for all of our Hamamatsu Cameras. Currently debugging on the Lightning, but confirmed it also happens on the BT-MesoSPIM. It happens when we set the exposure time. According to the documentation, the step size for incrementing the exposure time is quite small. 

![image](https://github.com/TheDeanLab/ASLM/assets/42547789/4fb79454-9337-4923-86e8-2c44fbb1a65b)

For the settings on the software, 100 ms exposure and 100 pixel rolling shutter readout, the exposure time passed to the `set_property_value` method in the `HamamatsuAPI` are the following is 0.00465766.  

[...]

PR #808, opened 2024-01-20
camera exposure time in lightsheet mode

annie-xd-wang:
#630: The error has been fixed for the Hamamatsu Orca Fire camera. It was due to inaccuracies in exposure time calculation for that specific camera. If the same error occurs on other machines, I'll need to read the specific documentation.


zacsimile:
Awesome! That makes sense. Specs for 1H can be found for the Flash (our current assumed default) at https://www.hamamatsu.com/content/dam/hamamatsu-photonics/sites/static/sys/en/manual/C13440-20CU_IM_En.pdf, the  Fusion BT at https://www.hamamatsu.com/content/dam/hamamatsu-photonics/sites/static/sys/en/manual/C15440-20UP,-20UP01_IM_En.pdf and the Lightning at https://www.hamamatsu.com/content/dam/hamamatsu-photonics/sites/static/sys/en/manual/C14120-20P_IM_En.pdf. 

It's worth noting that 1H changes depending on the camera's mode of operation and a few other settings. My original thought was we could just let it throw the error, since I believe this is from our side of the code (which has a fixed rounding tolerate =/= 1H, line 1043 of `HamamatsuAPI.py`), since the camera automatically then does its rounding, and we pull the true exposure time for whatever the current 1H is back from the camera itself. Effectively, if we divide by a particular 1H for whatever condition we most commonly use, it will work for that condition and then still throw this error if the camera's mode of operation (and possibly exposure time, depending on the camera) is changed.

zacsimile:
Looking at the code more closely, I don't fully understand why this fixes the error. Is `properly_value_step` as defined in line 1007 of `HamamatsuAPI.py` not equivalent to 1H? I think it should be.

annie-xd-wang:
@zacsimile, I agree with you that properly_value_step, defined in line 1007 of the HamamatsuAPI, is intended to round up 1H. However, it didn't behave that way the last time I worked with the camera. I will double-check the value of properly_value_step for exposure time the next time I connect to a camera.

zacsimile:
It didn't behave that way for me earlier, but I'm unclear as to why. I wondered if maybe it was returning 1H for a different exposure time. Maybe we need to set exposure time before line interval or vice versa to get the 1H to match. Or the logic is a bit different than I expect in the DCAM API.

annie-xd-wang:
The minimum exposure time we get from the camera is 1H. Casting the exposure time according to the step size causes accuracy errors.

zacsimile:
Nice find! Super weird that rounding to integer multiples the step size doesn't work, but glad it's working with the minimum now.

AdvancedImagingUTSW:
I quickly tried to test it on a Mac, and I got the following error when running it in the continuous mode. However, when I tried to get the same error again, it did not happen. Seems somewhat random. Was in continuous mode, 3 channels selected, hit stop, and it threw the error.

```
Exception in thread live signal:
  File "/opt/miniconda3/envs/navigate/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/opt/miniconda3/envs/navigate/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/S155475/Desktop/GitHub/navigate/src/navigate/model/model.py", line 1094, in run_live_acquisition
    self.signal_container.reset()
AttributeError: 'Model' object has no attribute 'signal_container'

```

src/navigate/controller/sub_controllers/camera_setting_controller.py
```diff
@@ -464,43 +463,23 @@ def calculate_physical_dimensions(self):
...
-            elif self.trigger_active == 3:
-                #  External Trigger Source
-                #  Synchronous Readout == 3
-                readout_time = exposure_time - ((vn / 2) * h + 5 * h)
+            readout_time = self.camera_setting_dict["readout_time"]
 
         elif sensor_mode == "Light-Sheet":
             #  Progressive sensor mode operation
-            readout_time = exposure_time - 1 / (exposure_time + (vn + 10) * h)
+            readout_time = 0
```
AdvancedImagingUTSW, line 482: Is this always the case? What happens if the exposure time is very short (e.g., 1 ms)?

src/navigate/model/device_startup_functions.py
```diff
@@ -1240,7 +1240,11 @@ def load_devices(configuration, is_synthetic=False, plugin_devices={}) -> dict:
                 )
                 # if the serial number is with leading zeros, the yaml reader will convert it to a octal number
                 if camera_serial_number.startswith("0"):
-                    devices["camera"][build_ref_name("_", device["type"], int(camera_serial_number, 8))] = camera
+                    try:
+                        oct_num = int(camera_serial_number, 8)
+                        devices["camera"][build_ref_name("_", device["type"], oct_num)] = camera
```
AdvancedImagingUTSW, line 1245: I still need to test this.

AdvancedImagingUTSW (approved):

PR #816, opened 2024-02-01
Stable widgets

### LINKED ISSUE #565: Waveform Popup Controller - Remove Hard Coded Microscope Names
We should move this to the GUI portion of the configuration yaml file. Line 202...

```        if (
            self.resolution == "high"
            or self.resolution == "Nanoscale"
            or self.resolution == "CTASLMv1"
        ):
            precision = -4
            increment = 0.0001
        else:
```

### LINKED ISSUE #775: Missing settings `Warning` prompted for no good reason...
I think this happens if the `set start pos/foc` and` set end pos/foc` buttons are pressed at the same location, in `continuous scan` mode. It seems to not like the `# slices` entry to be set to 0, which should still be read-only anyways. But for live scan imaging, it should not throw this error. For a `z-stack`, it makes sense.

image

annie-xd-wang:


src/navigate/controller/sub_controllers/camera_setting_controller.py
```diff
@@ -438,18 +443,12 @@ def calculate_physical_dimensions(self):
...
         try:
-            x_pixel = float(self.roi_widgets["Width"].get(self.min_width))
-            y_pixel = float(self.roi_widgets["Height"].get(self.min_height))
+            x_pixel = float(self.roi_widgets["Width"].get())
+            y_pixel = float(self.roi_widgets["Height"].get())
         except ValueError as e:
-            logger.error(f"{e} similar to TclError")
             return
 
-        # Round to nearest step
```
AdvancedImagingUTSW, line 449: @zacsimile - This is to account for the minimum step size that a camera can accept, right? 

@annie-xd-wang - Still reading the code, but is this feature introduced elsewhere?

zacsimile: Yes, I added this for setting the ROI on the HamamatsuLightning. It is technically also addressed in the DCAM API get/set, but it's here as well to make the GUI provide instant feedback.

src/navigate/controller/sub_controllers/channel_setting_controller.py
```diff
@@ -413,3 +411,32 @@ def get_index(self, dropdown_name, value):
...
+
+    def verify_experiment_values(self):
+        """Verify channel settings and return warning info
+        
+        Returns
+        -------
+        string
+            Warning info
+        """
+        selected_channel_num = 0
```
AdvancedImagingUTSW, line 432: Warnings should be adjusted slightly. Lower laser power, exposure time, etc., than what?

src/navigate/controller/sub_controllers/waveform_popup_controller.py
```diff
@@ -198,33 +204,32 @@ def configure_widget_range(self):
...
-            precision = -3
-            increment = 0.001
-
-        laser_min = self.configuration_controller.remote_focus_dict["hardware"]["min"]
-        laser_max = self.configuration_controller.remote_focus_dict["hardware"]["max"]
+        self.laser_min = self.configuration_controller.remote_focus_dict["hardware"]["min"]
+        self.laser_max = self.configuration_controller.remote_focus_dict["hardware"]["max"]
+        
+        precision = int(self.configuration_controller.remote_focus_dict["hardware"].get("precisition", 0))
+        increment = int(self.configuration_controller.remote_focus_dict["hardware"].get("step", 0))
```
AdvancedImagingUTSW, line 211: `prercisition` is misspelled. Could be pulling 0 by default since that entry won't likely be found in the remote_focus_dict...

src/navigate/view/custom_widgets/validation.py
```diff
@@ -337,24 +341,14 @@ def undo(self, event):
...
             # Get the redo value
             value = self.undo_history.pop()
-            if self.redo_history and self.redo_history[-1] == value:
+            if self.redo_history and self.redo_history[-1] == self.get():
                 pass
             else:
-                self.redo_history.append(value)
+                self.redo_history.append(self.get())
             if len(self.redo_history) > 3:
                 self.redo_history.pop(0)
```
AdvancedImagingUTSW, line 351: @zacsimile - can you double check this? I know you worked on `validation` quite a bit...

src/navigate/view/main_window_content/channels_tab.py
```diff
@@ -316,6 +317,8 @@ def populate_frame(self, channels):
             self.interval_spins[num].grid(
                 row=num + 1, column=0, sticky=tk.NSEW, padx=1, pady=1
             )
+            hover = Hover(self.interval_spins[num], text="Not Implemented", type="free")
+            hover.setdescription(text="Not Implemented")
```
AdvancedImagingUTSW, line 321: Should make sure this works on a Windows machine. Hover is showing up as empty for me. 

image

src/navigate/controller/sub_controllers/channel_setting_controller.py
AdvancedImagingUTSW: Specifically:

"Warning: Laser power below configured threshold. Please adjust to meet or exceed the specified minimum in the configuration.yaml."

"Warning: Laser power exceeds configured maximum. Please adjust to meet or be below the specified maximum in the configuration.yaml."

This could be the same for the other warnings too.

src/navigate/controller/sub_controllers/camera_setting_controller.py
AdvancedImagingUTSW: @annie-xd-wang - We should probably leave this, unless there is a good way to have the camera class, in this case DCAM, provide instant feedback to the GUI. Otherwise, it seems as if the user may think the camera is one width, but when the acquisition actually begins, it will be something different.

AdvancedImagingUTSW:
Changes look great. Once we address a few of the things discussed above, we can accept, merge, and push a new version to pypi.

src/navigate/view/custom_widgets/validation.py
zacsimile: I didn't read the code thoroughly but I just tried it out on my mac and undo doesn't work for me on the stage widgets anymore.

src/navigate/view/custom_widgets/validation.py
```diff
@@ -212,6 +213,8 @@ def _validate(self, proposed, current, char, event, index, action):
                 index=index,
                 action=action,
             )
+            if valid:
```
zacsimile, line 217: Maybe needs to be unindented?

src/navigate/view/custom_widgets/validation.py
zacsimile: Not sure if it's the logic or the keybindings that aren't working.

src/navigate/controller/sub_controllers/channel_setting_controller.py
annie-xd-wang: fixed

src/navigate/controller/sub_controllers/waveform_popup_controller.py
annie-xd-wang: fixed

src/navigate/view/custom_widgets/validation.py
annie-xd-wang: It's okay. Focusout and key events employ different strategies for handling inputs.

src/navigate/view/custom_widgets/validation.py
annie-xd-wang: It has been fixed. Every input to an Entry will surprisingly trigger one Focusout event after one key event, regardless of whether the widget is still focused. The same thing happens if we directly set the entry value. We have to mark whether the Focusout event is a valid one.

src/navigate/view/main_window_content/channels_tab.py
annie-xd-wang: It displays correctly on my machine. I suspect the issue is related to the font color.

src/navigate/controller/sub_controllers/camera_setting_controller.py
AdvancedImagingUTSW: Currently, the software will take a value that is not perfect, but when we hit `Acquire` it checks the value, and then sets it to the next closest step size that is valid. Seems probably good enough since it is a very minor change, and the software does eventually get to the right value.

src/navigate/view/custom_widgets/validation.py
AdvancedImagingUTSW: If I toggle an entry so that it is empty, and then leave it, get this error:

```
  File "/opt/miniconda3/envs/navigate/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
    return self.func(*args)
  File "/Users/S155475/Desktop/GitHub/navigate/src/navigate/view/custom_widgets/validation.py", line 746, in _validate
    self.undo_history.pop()
IndexError: pop from empty list
	
```

self.undo_history should be populated with the first value, or check to see if it is empty?

AdvancedImagingUTSW: Make it empty, focus out, go back, focus out, throws the error.

AdvancedImagingUTSW: Partially fixed with the latest push. 

However, if we focus out with it empty, it should revert to the last value.

src/navigate/view/main_window_content/channels_tab.py
AdvancedImagingUTSW: If it works on your machine, I'm good with it. I will confirm in lab when I have a second.

src/navigate/view/custom_widgets/validation.py
annie-xd-wang: Fixed. Tkinter will trigger a Focusout event if a key event occurs. However, the Focusout event won't be triggered when deleting the value to empty.

src/navigate/view/main_window_content/channels_tab.py
AdvancedImagingUTSW: Now also works on Mac.

AdvancedImagingUTSW (approved):
I've tried to evaluate everything and it appear to be working to the best of my knowledge.

PR #822, opened 2024-02-06
Stable widgets

### LINKED ISSUE #821: Validated Entry (Presumably)
If more than one item is selected, you cannot delete. Any values that are introduced go to the leading position strangely. Cannot easily change things in the save acquisition dialog.

annie-xd-wang:


AdvancedImagingUTSW (approved):
Much easier to work with on the microscope. Thank you @annie-xd-wang

PR #824, opened 2024-02-08
818 fatal not a git repository or any of the parent directories git

### LINKED ISSUE #818: fatal: not a git repository (or any of the parent directories): .git
When downloaded from PyPi, it is no longer a git repository. At some point, we should make sure that the git hash call works properly.

```
fatal: not a git repository (or any of the parent directories): .git
```

AdvancedImagingUTSW:
I tried to address the unnecessary error, as well as the comment to have the `__version__` be dynamic. 

It prioritizes the git hash, if we are in a git repository, since this is more specific. 

If we are not in a git repository (e.g., downloaded form PyPI), then it pulls the version number from the VERSION file.

Wanted to make sure this aligned with your vision of it being dynamic.

### DIFF
--- a/src/navigate/_commit.py
+++ b/src/navigate/_commit.py
@@ -41,31 +41,64 @@ import subprocess
 
 
 def get_git_revision_hash() -> str:
+    """ Get the git commit hash of the current repository.
+
+    This function will return the git commit hash of the current repository, or None if
+    the current directory is not a git repository.
+
+    Returns:
+    --------
+    commit_hash : str
+        The git commit hash of the current repository, or None if the current directory
+        is not a git repository
     """
-    TODO: This should not be dynamic. A packaged version will still have an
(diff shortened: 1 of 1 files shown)


src/navigate/_commit.py
```diff
@@ -41,25 +41,64 @@
...
+
+__commit__ = get_git_revision_hash()
+if __commit__ is None:
+    __commit__ = get_version_from_file()
 
-try:
-    __commit__ = get_git_revision_hash()
-except Exception:  # noqa
-    __commit__ = "unknown"
+print(f"Commit: {__commit__}")
```
zacsimile: This prints quite a few times. I might remove.

zacsimile:
Yep, seems reasonable.

zacsimile (dismissed):

PR #831, opened 2024-02-15
830 continuous acquisition with multiple channels

### LINKED ISSUE #830: Continuous Acquisition with Multiple Channels
**Describe the bug**
When multiple channels are selected, and the microscope is operated in the continuous scan mode, it toggles between the channels once and then stops and only displays a single channel. If CH0 and CH1 are selected, it will display CH0, then CH1, then only CH1. In other words, it stays on the highest channel.

**Expected behavior**
Should continuously toggle between the channels with any offset specified by the GUI.

**Traceback**
Nothing in traceback.

annie-xd-wang:


AdvancedImagingUTSW (approved):
This fixes it. Thank you @annie-xd-wang

PR #834, opened 2024-02-26
Update channels_tab.py

### LINKED ISSUE #833: Tab of laser power control
I always need to be careful whether I enter 10% or 100% of the laser power. Because the size of the tab to enter the laser power is simply just too small. 

Can you make it larger? 
Thanks.

image

AdvancedImagingUTSW:


AdvancedImagingUTSW:
@bojuichang - Why don't you get a copy of this pull request on the machine and see if you are happy. Let me know.

PR #848, opened 2024-03-19
842 configurationyaml configurator gui

### LINKED ISSUE #842: Configuration.yaml Configurator GUI
Misc. Ideas.
Would love feedback. Very early stage. I am just concerned that the configuration.yaml file is the most fragile part of the software, which could slow adoption for non-experts. Would like a way to eliminate this problem.

Should the program be launched from a parser argument (e.g., navigate -config) or as a separate program?

Should the program be able to load and modify and existing configuration.yaml file?

How do we store what variables need to be changed for each device?

How do we store what devices are available (__init__.py?)?

It should also be able to configure multiple microscope objects.

By default, if the configuration.yaml file is not found, this program should automatically be launched. 

Maybe have a menu option so that if the software is launched in the synthetic hardware mode, that they can access this configurator as well. 

We want them to be able to create a new configuration, as well as modify an existing one.

I originally thought that it could be a tabbed frame for each microscope, with a GUI option to add more microscopes. Within each microscope, would be another tabbed frame for each device, etc.

For example, one idea is to have a known dictionary associated with each device that can be imported independently. For the Hamamatsu Cameras, could be something like:
```
# Logger Setup
p = __name__.split(".")[1]
logger = logging.getLogger(p)

required_params = {
    "serial_number": "str",
    "lightsheet_rolling_shutter_width": "int",
    "defect_correct_mode": "float",
    "delay_percent": "int",
    "pulse_percent": "int",
    "x_pixels_step": "int",
    "y_pixels_step": "int",
    "x_pixels_min": "int",
    "y_pixels_min": "int",
    "exposure_time_range": "dict",
    "flip_x": "bool",
    "flip_y": "bool",
}
```

There will be some odd things. Like axes_mapping for the stages will probably require a creative solution.

Other items may be complicated. For example, let's think about the lasers. Analog and digital modulation is defined independently. So, the analog input could be a list of options (`["Synthetic Laser", "NI Laser"]`). And then some specific options will then be populated depending upon that option? Or maybe we just show everything (e.g., max voltage, min voltage, channel, wavelength) for both? Or maybe make some read only rather than populate them based upon the selection? Also, some microscopes may have 4 lasers, others 7, and we need to be able to dynamically select the number of lasers.

AdvancedImagingUTSW:
This is a partial PR. I didn't want it to get too big...

- New command line argument provided: -c --configurator to launch the configuration assistant.
- `configurator.py` is the controller.
- `configurator_application_window.py` is the view.
- User can specify the number of microscopes they would like in their configuration. 
- User selects each device type, which is pulled from the `__init__.py` file.

TODO:
- Still needs a user entry for them to name each microscope.
- Currently, if they change the number of microscopes midway through the configuration, It think it will delete everything. Need a better way. For example, if they specify 2 microscopes, do a partial configuration, then specify 3, it should keep the original 2 not delete them.
- Add two columns to the device window after they select the device type. One will be for required parameters, and the other will be for optional parameters.
- No logic for creating YAML yet provided.


annie-xd-wang:
Do not test it now. The functions for generating configurations for microscopes are now working. They do not generate the hardware header section. I believe we should remove the header section so that the configuration does not contain redundant information. I will update the device_startup_functions later to not use the hardware header information.

PR #852, opened 2024-04-14
Additional Camera Support

AdvancedImagingUTSW:
Will perform a couple of local tests locally at UTSW on different hardware and confirm that we are getting the right values.


AdvancedImagingUTSW (commented):
Confirmed to work.

PR #859, opened 2024-04-24
fix bug: Error message when saving binned images

### LINKED ISSUE #858: Error message when saving binned images
When saving binned images (single acquisition), an error message shows up that there is insufficient disk space. Issue reproduced with Hamamatsu and Photometrics camera on two setups.
Stephan

annie-xd-wang:


src/navigate/model/metadata_sources/metadata.py
```diff
@@ -169,11 +169,12 @@ def set_shape_from_configuration_experiment(self) -> None:
         self.dz = float(abs(state["step_size"]))
         self.dt = float(state["timepoint_interval"])
 
+        # TODO: do we need to update the XML meta data accordingly?
```
zacsimile, line 172: I don't think so. They use `self.shape_x`.

AdvancedImagingUTSW:
@zacsimile - I wasn't confident enough with your line comment to proceed with approving the change. If you are happy with it, can you remove the comment about the metadata, and approve the PR?

zacsimile:
I was just responding to the code comment, saying we don't need to do anything else. The PR is fine.

zacsimile (approved):

PR #863, opened 2024-04-25
Add OME-Zarr Data Source

### LINKED ISSUE #766: Add Zarr data source
Add an OME-Zarr data source.

zacsimile:
This PR...

* Establishes a generalized pyramidal data source for multi-resolution writing.
* Adds an OME-Zarr data source/writer to `navigate` (closes #766).
* Includes a fix for `uint16` writing in BDV H5 files as described here: https://github.com/bigdataviewer/bigdataviewer-core/issues/102.

The OME-Zarr writing seems to work. I can open it in a notebook and it looks to me like it satisfies the minimum requirements at https://ngff.openmicroscopy.org/0.4/. I can open the Zarr with File > Import in ImageJ. However, even though the path and shape discovery works, I cannot open the Zarr images with Plugins > BDV > HDF5/N5/Zarr/OME-NGFF Viewer. I'm not sure if this is due to a bug in my writer, lack of support for NGFF version 0.4 in this plugin, or something else. Does anyone with Zarr experience (@dpshepherd ?) have any ideas as to what is going on?

*** PLEASE CHECK THIS WORKS ON ACTUAL MICROSCOPES BEFORE MERGING. ***

Future: Write speed is not optimized for N5 or Zarr. We should thread the writing, possibly with a `Synchronizer` (https://zarr.readthedocs.io/en/stable/api/sync.html), to achieve faster speeds. Alternatively, we could decline to write the subsampled versions of the image for now.

Future: The reader is half-implemented and won't read the metadata at the moment.

### DIFF
--- a/src/navigate/model/data_sources/__init__.py
+++ b/src/navigate/model/data_sources/__init__.py
@@ -1,6 +1,6 @@
 """ File type specific data sources. """
 
-FILE_TYPES = ["TIFF", "OME-TIFF", "H5", "N5"]
+FILE_TYPES = ["TIFF", "OME-TIFF", "H5", "N5", "OME-Zarr"]
 
 
 def get_data_source(file_type):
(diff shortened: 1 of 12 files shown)


AdvancedImagingUTSW (dismissed):
Will push my few formatting changes, etc.

dpshepherd:
I've used this to validate that a ome-zarr is spec v.4 compliant :
https://janeliascicomp.github.io/pydantic-ome-ngff/#data-validation

We don't use the Fiji import for ome-zarr, so I'm not sure about what might be different there. My feeling is if it passes the pydantic tool's test, then it is good to go.

PR #865, opened 2024-04-25
Qi2lab - ASI MS2000

nng-thienphu:
Adds support for ASI MS2000 apis and stage, tested successfully with the GUI and hardware operation.


AdvancedImagingUTSW:
We are talking this device, right?

Can it control a filter wheel @nng-thienphu ?

src/navigate/model/device_startup_functions.py
```diff
@@ -488,6 +488,37 @@ def load_stages(configuration, is_synthetic=False, plugin_devices={}):
                         exception=TigerException,
                     )
                 )
+        
+        elif stage_type == "MS2000" and platform.system() == "Windows":
+            """ Filter wheel can be controlled from the same Tiger Controller. If
+            so, then we will load this as a shared device. If not, we will create the
+            connection to the Tiger Controller.
+            """
+            filter_wheel = configuration["configuration"]["hardware"]["filter_wheel"][
```
AdvancedImagingUTSW, line 497: I'm tempted to remove this, but then again, the MS2000 can control the CRISP, which could someday be an `AutofocusingDevice`?

AdvancedImagingUTSW: I now realize that we did the same thing with the MFC2000 @ConnorHoyUTSW 

Perhaps we can leave it, but I will insert a TODO in the code.

AdvancedImagingUTSW (approved):

PR #874, opened 2024-05-03
Update cva_conpro.py

AdvancedImagingUTSW:
@mehr0096 - We should avoid having all of these print statements pushed to develop.


mehr0096:
Thanks, I removed more print statements, unused variables, commented out code.

PR #887, opened 2024-05-17
Update file format documentation

zacsimile:


AdvancedImagingUTSW (approved):
I see you switched mean for median. We should probably consider doing this in the manuscript as well.

zacsimile:
Did it in the manuscript before I submitted this PR :).

PR #891, opened 2024-05-21
Add stage flip flag documentation

zacsimile:
Still missing camera flip flag documentation. @annie-xd-wang do you remember why we need both camera and stage flip flags?


annie-xd-wang:
Yes, I added the flip flags to the stage.rst and missed this file. Thank you for adding it.

AdvancedImagingUTSW (approved):

PR #892, opened 2024-05-21
Stage event bindings

### LINKED ISSUE #882: Stage controller initiates a move before the user finishes typing
This issue concerns the GUI stage control window. When typing a position in any of the X, Y, Z stage text entry boxes, the stage begins to move to the entered value before the user finishes typing and changes focus by leaving the text box. For example, if I pause for a moment at 10 while typing 1000, the stage starts moving to 10.

While I don't anticipate this being a frequent occurrence, it does introduce the risk of a collision. Could the text box be configured to update the stage only when the user leaves the text box?

annie-xd-wang:


AdvancedImagingUTSW (approved):
Tested with print statement in `move_axis_absolute`, worked perfectly.

PR #903, opened 2024-05-24
TTL controlled Filter wheel using NI DAQ

SJShep:
Added a NI DAQ filter wheel class that works with the NI DAQ class. There is an additional function in *ni_daq.py* to send a single digital pulse. We create the connection in *microscope.py* under shared devices. In *device_startup_functions.py* we included the new filter wheel class and also rearranged device loading to load the DAQ before loading filter wheels. 

Configuration snippit: 
```
filter_wheel:
      hardware:
        name: filter_wheel
        type: NI  
        wheel_number: 1
      filter_wheel_delay: 0.050 # in seconds
      available_filters:
        473nm: Dev2/port0/line1
        561nm: Dev2/port0/line3
        638nm: Dev2/port0/line5
        Empty: Dev2/port0/line7
```


annie-xd-wang:
I've made some changes to this PR, but unfortunately, I don't have permission to push commits to it. So I created another branch [(updated_ni_filter_wheel)](https://github.com/TheDeanLab/navigate/tree/updated_ni_filter_wheel) based on this PR.

SJShep:
> I've made some changes to this PR, but unfortunately, I don't have permission to push commits to it. So I created another branch [(updated_ni_filter_wheel)](https://github.com/TheDeanLab/navigate/tree/updated_ni_filter_wheel) based on this PR.

No problem, I will let you know if we have any issues with the changes.

PR #919, opened 2024-06-10
Multiple filter wheels

annie-xd-wang:


AdvancedImagingUTSW (approved):
Confirmed to work on microscopes with:
Sutter Instruments - Single Filter Wheel
ASI - Single Filter Wheel
ASI - Multiple Filter Wheels

I don't anticipate problems with other filter wheels, but let us know if we do have any problems @conorhughmcfadden @SJShep

PR #933, opened 2024-07-06
Refactor naming

AdvancedImagingUTSW:


AdvancedImagingUTSW:
When you get back, hopefully this is not too annoying. My goal was to make some of our imports shorter. If you have the pre-commit linter operating, then many of our imports throw errors. We could consider something similar for the sub-controllers. Given that they are in a folder titled sub-controller, having controller in their module names is also redundant.

AdvancedImagingUTSW:
Tested on CT-ASLM-V2 and BT-MesoSPIM. Works as expected.

PR #937, opened 2024-07-12
New feature: WaitForExternalTrigger

conorhughmcfadden:
Put this feature in your feature list to wait for an external digital trigger before moving on to the next block.

Example feature list which worked:

`[{"name": PrepareNextChannel,},({"name": WaitForExternalTrigger,"args": ("/PCIe-6738/PFI4",-1,),},({"name": MoveToNextPositionInMultiPositionTable,},{"name": ZStackAcquisition,"args": (True,True,),},{"name": WaitToContinue,},{"name": LoopByCount,"args": ("experiment.MicroscopeState.multiposition_count",),},),{"name": LoopByCount,"args": (5,),},),]`

Option to set a timeout to break the loop if no trigger is incoming.

Tested with a function generator hooked up to PFI4 and it seemed to work well.


conorhughmcfadden:
I found a bug...

In the proposed feature list using multiposition above, the stage will not go back to the first position after the first iteration. It just repeatedly stacks the last position.

I'll investigate why and report back.

conorhughmcfadden:
> I found a bug...
> 
> In the proposed feature list using multiposition above, the stage will not go back to the first position after the first iteration. It just repeatedly stacks the last position.
> 
> I'll investigate why and report back.

I found the issue! There appears to have been a bug in MoveToNextPositionInMultiPositionTable. At least when I ran this feature list without my new triggering block, just a normal loop, I got the same issue:

`[({"name": PrepareNextChannel, },({"name": MoveToNextPositionInMultiPositionTable, },{"name": ZStackAcquisition, "args": (False,False,"z-stack",),},{"name": WaitToContinue, },{"name": LoopByCount, "args": ("experiment.MicroscopeState.multiposition_count",),},),{"name": LoopByCount, "args": (3,),},),]`

It turns out current_idx needs to be reset in the signal_func in order for it to work with LoopByCount. Making this bug fix allows it to return to the original position using a nested loop. I tested it with my WaitForExternalTrigger and it works properly now!

AdvancedImagingUTSW:
This is a great addition. Before we merge, I think it would be a good idea to have the calls to the data acquisition card go through the device abstraction layer. This way, should we add additional ways to receive external signals (e.g., Tiger Controller, [TriggerScope](https://nvladimus.medium.com/instrument-review-triggerscope-4-0-5a3768e0ab73)), it will be more modular.

src/navigate/model/features/common_features.py
```diff
@@ -199,6 +200,76 @@ def data_func(self, frame_ids):
...
+        }
+
+    def signal_func(self):
+
+        # Pause the data thread to prevent camera timeout
+        self.model.pause_data_thread()
+
+        # Create a digital input task and wait until either a trigger is detected,
+        # or the timeout is exceeded. If timeout < 0, wait forever...
+        self.task = nidaqmx.Task('WaitDigitalEdge')
```
AdvancedImagingUTSW: self.task = self.model.active_microscope.daq.create_digital_edge()...

This way if we add a different daq in the future, from another manufacturer, we do not have to rewrite the feature.

src/navigate/model/features/common_features.py
```diff
@@ -624,6 +695,11 @@ def signal_func(self):
         self.model.move_stage(abs_pos_dict, wait_until_done=True)
 
         self.model.logger.debug("MoveToNextPositionInMultiPosition: move done")
+        
+        # Make sure to go back to the beginning if using LoopByCount
+        if self.current_idx == self.position_count:
```
AdvancedImagingUTSW: According to @annie-xd-wang, the multi position feature does not reset the index for the position in the table after it loops. Might be better to update there. She will take a look at it.

AdvancedImagingUTSW: @annie-xd-wang will adjust the code, and then @conorhughmcfadden can test it again on his system, and I will test it on one of our rigs too. If all goes well, we will merge it.

annie-xd-wang: @conorhughmcfadden has fixed it.

src/navigate/model/features/common_features.py
annie-xd-wang: The task is moved to daq_ni.py. Please test the code.

annie-xd-wang (dismissed):

src/navigate/model/features/common_features.py
conorhughmcfadden: I tested it on my system and Annie's changes work well! There was a minor bug fix but just a typo. I tested the following feature:

`[({"name": WaitForExternalTrigger,"args": ("/PCIe-6738/PFI4",-1,),},{"name": PrepareNextChannel,},({"name": MoveToNextPositionInMultiPositionTable,},{"name": ZStackAcquisition,"args": (False,False,"z-stack",),},{"name": WaitToContinue,},{"name": LoopByCount,"args": ("experiment.MicroscopeState.multiposition_count",),},),{"name": LoopByCount,"args": (3,),},),]`

Works the same as before!

AdvancedImagingUTSW (approved):

PR #939, opened 2024-07-13
Refactor plugins_model

AdvancedImagingUTSW:
In an attempt to write unit tests for the plugins_model, I found the function load_plugins too large to effectively test. So, I tried to break it up into smaller methods, which could then be tested independently.

Also found duplicated code in the plugins_controller, so I created a method of that section of code, and placed it in config.py. Both the plugins model and the plugins controller now load this function...

Will test before moving on to write unit tests...


AdvancedImagingUTSW:
@annie-xd-wang - Before you do too much work on making the plugins pip installable, could you review and test this PR?

annie-xd-wang (approved):

PR #940, opened 2024-07-16
Rename sub-controllers

AdvancedImagingUTSW:
Again, making an effort to make our imports less insane. Could also consider consolidating sub-controllers by function - e.g., anything that changes the camera view for example, could be in a single module.


AdvancedImagingUTSW:
I believe that everything is working properly, but I figured it wouldn't hurt to have a proper second opinion @annie-xd-wang

annie-xd-wang (approved):

PR #949, opened 2024-07-22
Add stage offset in the MoveToNextPositionInMultiPositionTable feature

annie-xd-wang:


AdvancedImagingUTSW:
Background: The software provides for an offset to be applied when switching between different microscopes. For Jinlong's system, you could find a feature in the low-res side, then switch the microscope, and the coordinates from the configuration.yaml file would be pulled and applied to the stages:
```
      x_offset: 0
      y_offset: 0
      z_offset: 0
      f_offset: 0
      theta_offset: 0
```
However, this correction was only applied once (e.g., to your current stage position). If features were added to the multi-position table in one mode, and then you switched to the other mode, the offset would not have been applied to the multi-position table. Now, the multi-position table logs which mode the positions were added in, and any subsequent move in a different mode will apply the offset provided in the configuration file.

AdvancedImagingUTSW (approved):

AdvancedImagingUTSW:
Tested by @JinlongL. Some unit tests would be valuable in the future, and we need to document how to properly use these features.

PR #951, opened 2024-07-23
Communicating from model to controller

annie-xd-wang:
Register a custom event in a controller or plugin_controller in the following way:

    @property
    def custom_events(self):
        return {
            "autofocus": self.display_plot
        }

Send data back in a feature to the controller:

      self.model.event_queue.put(("autofocus", [fit_data, True, False]))

You can read model/features/autofocus.py and controller/sub_controllers/autofocus.py as an example.

### DIFF
--- a/src/navigate/model/model.py
+++ b/src/navigate/model/model.py
@@ -512,15 +512,6 @@ class Model:
                 "is_save"
             ]
 
-            # If multiposition is selected, verify that it is not empty.
-            if self.configuration["experiment"]["MicroscopeState"]["is_multiposition"]:
-                if len(self.configuration["experiment"]["MultiPositions"]) == 0:
-                    # Update the view and override the settings.
-                    self.event_queue.put(("disable_multiposition", None))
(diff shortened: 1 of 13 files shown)


src/navigate/model/model.py
```diff
@@ -512,15 +512,6 @@ def run_command(self, command, *args, **kwargs):
                 "is_save"
             ]
 
-            # If multiposition is selected, verify that it is not empty.
```
AdvancedImagingUTSW, line 515: Why do we remove this logic? Previously, if the multi position button was selected, but the multi position list was empty, it would crash the program.

AdvancedImagingUTSW: Okay, I see now. You are handling this directly in the `controller` which is logically cleaner since it avoids the need to put something in the `event_queue`.

AdvancedImagingUTSW (approved):
I'm still a bit perplexed by the logic here, but it seems as if the GUI sub-controller identifies all of the events, and then registered these events with the controller.

PR #960, opened 2024-08-07
Fix errors in displaying with multiple cameras

annie-xd-wang:


AdvancedImagingUTSW:
@JinlongL Can you test this out on the TIRF microscope with two cameras working? Check the full camera size, as well as a sub-ROI (e.g., 512x512)?

JinlongL:
Thank you. This branch is working. So we can use both cameras and change binning, FOV.
________________________________

AdvancedImagingUTSW (approved):

PR #965, opened 2024-08-14
445 arbitrary ROI

### LINKED ISSUE #445: Add ability to set arbitrary ROI
By default, we cut about the middle of the FOV. However, it is perfectly reasonable to create an ROI from, say, `(500, 430)` to `(700, 490)` or something similarly random. We should implement this in the logic/GUI rather than set the width/height and ROI center (which is equivalent, but currently disabled).

annie-xd-wang:


AdvancedImagingUTSW:
Small bug found when manually adjusting ROI Boundary widgets...

```
  File "C:\Users\Dean-Lab\Desktop\navigate\src\navigate\controller\sub_controllers\camera_settings.py", line 481, in update_fov
    if value < 0 or value > max_value or value % step_value:
TypeError: '<' not supported between instances of 'str' and 'int'
```

AdvancedImagingUTSW (approved):
Very nice. I have confirmed that it works on one microscope.

PR #993, opened 2024-09-19
Update asi_tiger_controller.py

AdvancedImagingUTSW:
On John's system, there are other types of cards that are not included with our other Tiger Controller builds. This includes piezos, for example, which do not recognize some of the commands we use to set the speed as a percent of the max speed. This fix now goes through all of the stages and removes those that do not qualify.

https://www.asiimaging.com/docs/products/serial_commands#commandbuild_bu


AdvancedImagingUTSW:
We should test this on the BT-MesoSPIM before proceeding with the merge...

AdvancedImagingUTSW:
Confirmed to work on BT-MesoSPIM.

Johnhaug223 (approved):

PR #1018, opened 2024-10-16
Close ni tasks

AdvancedImagingUTSW:
Tried to create a more uniform way for closing tasks so that our console isn't cluttered upon exit.

There remains one traceback from the threading library.

If we replace `raise SystemExit()` with `os._exit(1)` in the following code this is eliminated completely:

```
        if event == "exception":
            if os.getenv("GITHUB_ACTIONS") == "true":
                return

            # Silence traceback to avoid printing to console.
            sys.tracebacklimit = 0
            raise SystemExit()
        return self.localtrace
```

I was hesitant to make this change because this method apparently does not invoke the normal Python cleanup routines (like flushing I/O buffers, calling finally blocks, etc.).

### DIFF
--- a/src/navigate/model/devices/camera/hamamatsu.py
+++ b/src/navigate/model/devices/camera/hamamatsu.py
@@ -143,7 +143,7 @@ class HamamatsuBase(CameraBase):
 
     def __del__(self):
         """Delete HamamatsuOrca class."""
-        pass
+        self.close_camera()
 
     @property
     def serial_number(self):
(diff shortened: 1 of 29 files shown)


src/navigate/model/devices/camera/hamamatsu.py
```diff
@@ -143,7 +143,7 @@ def __str__(self):
 
     def __del__(self):
         """Delete HamamatsuOrca class."""
-        pass
+        self.close_camera()
```
annie-xd-wang: Calling close_camera() here causes a DCAM error.

annie-xd-wang (approved):

PR #1030, opened 2024-11-08
Speed up loading multi-positions

annie-xd-wang:


src/navigate/controller/sub_controllers/camera_view.py
```diff
@@ -526,8 +526,8 @@ def initialize_non_live_display(self, microscope_state, camera_parameters):
         self.slice_index = 0
         self.image_mode = microscope_state["image_mode"]
         self.stack_cycling_mode = microscope_state["stack_cycling_mode"]
-        self.number_of_channels = int(microscope_state["selected_channels"])
         self.get_selected_channels(microscope_state)
+        self.number_of_channels = len(self.selected_channels)
```
AdvancedImagingUTSW, line 530: This causes an error when we switch to the non-live view and then adjust the slider position. To test it, I did a z-stack with only CH3 selected.

```
Exception in Tkinter callback
  File "/opt/miniconda3/envs/navigate/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
    return self.func(*args)
  File "/Users/S155475/Desktop/GitHub/navigate/src/navigate/controller/sub_controllers/camera_view.py", line 1174, in slider_update
    image = self.spooled_images.load_image(
  File "/Users/S155475/Desktop/GitHub/navigate/src/navigate/controller/sub_controllers/camera_view.py", line 1794, in load_image
    self.temp_files[channel].seek(start_idx)
KeyError: 2
```

src/navigate/model/features/common_features.py
```diff
@@ -451,31 +459,40 @@ def __init__(self, model, steps=1):
         self.model = model
 
         #: bool: A boolean value indicating whether to step by frame or by step.
-        self.step_by_frame = True
+        self.step_by_frame = False if type(steps) is str else True
```
AdvancedImagingUTSW, line 462: We should have @conorhughmcfadden check this PR.

AdvancedImagingUTSW (commented):
Looks great. We just need to make sure that the changes in the `WaitForExternalTrigger` do not disrupt @conorhughmcfadden's workflows. Also, should fix the slider, which probably would have been caught if I had created unit tests for it like I should have.

conorhughmcfadden:
I'm getting a frame timeout error when using the Photometrics camera on this PR. This is just running a normal multiposition feature list:

`[{"name": PrepareNextChannel,},({"name": MoveToNextPositionInMultiPositionTable,},{"name": ZStackAcquisition,"args": (True,True,),},{"name": WaitToContinue,},{"name": LoopByCount,"args": ("experiment.MicroscopeState.multiposition_count",),},),]`

The feature will run and I'll get to the final position, and the camera will hang throwing several frame timeout errors:

`frame timeout. Verify timeout exceeds exposure time. If applicable, check external trigger source.`

I've seen this happen occasionally with the Photometrics camera, but I did compare to develop and everything was fine, so I suspect it might be something with this PR.

Here is the `model_debug` when I recreated the error:

[model_debug.log](https://github.com/user-attachments/files/17708280/model_debug.log)

annie-xd-wang:
The parameter "multiposition_count" is removed from experiment.yml, accordingly, the parameter for LoopByCount needs to be set as "positions". Please update the feature pipeline as follows: [{"name": PrepareNextChannel,},({"name": MoveToNextPositionInMultiPositionTable,},{"name": ZStackAcquisition,"args": (True,True,),},{"name": WaitToContinue,},{"name": LoopByCount,"args": ("positions",),},),]

conorhughmcfadden:
I'm using the correct feature syntax now with `"positions"` but still get the same error:

[model_debug.log](https://github.com/user-attachments/files/17708410/model_debug.log)

I suspect the Photometrics is causing problems as I had similar issues with the AO feature when used in a feature list. The camera would just hang at the end. Does everything work alright with the Hamamatsu?

conorhughmcfadden:
Annie's new LoopByCount fix seems to have solved the problem. I tested WaitForExternalTrigger and it works as well.

conorhughmcfadden (commented):
I've tested the changes to LoopByCount with my workflow and it now runs without errors.

conorhughmcfadden (approved):

PR #1033, opened 2024-11-15
update ChangeResolution and SetCameraParameters

annie-xd-wang:
add parameter verification


AdvancedImagingUTSW (approved):
I find it a bit weird that the logger isn't working in the ChangeResolution feature. I feel like that is something worth looking into.

PR #1045, opened 2024-12-13
Modified for running in Linux and New stage driver

SJShep:
Made changes for running Navigate in Ubuntu 22.04 and added a stage driver for the KST101 in Linux.
There are also some changes scattered in to the the tiling, stage and NI related functions that are specific to either Linux or our setup. For example, we made some **bad** changes to hard code the direction of the  focus stage. For these changes a long term solution is needed.  Another example is in the NI daq code, the *register_done_event* function throws an error. The changes should be well commented.


src/navigate/controller/sub_controllers/tiling.py
```diff
@@ -266,26 +266,19 @@ def set_table(self):
         y_stop = float(self.variables["y_end"].get())
         y_tiles = int(self.variables["y_tiles"].get())
 
+        # NOTE: Removed shifting by the origin becuase, it was not clear how to set the origin.
```
AdvancedImagingUTSW, line 269: We should discuss this. I'm hesitant to make the change immediately, but we can consider a better way to communicate the origin.

src/navigate/model/device_startup_functions.py
```diff
@@ -511,8 +511,18 @@ def load_stages(
                     exception=TLFTDICommunicationError,
                 )
             )
-
-        elif stage_type == "KST101":
+        elif stage_type == "KINESIS" and platform.system() == "Linux":
```
AdvancedImagingUTSW, line 512: This is now obsolete with the latest PR.

src/navigate/model/device_startup_functions.py
```diff
@@ -705,7 +711,10 @@ def start_stage(
         from navigate.model.devices.stages.tl_kcube_inertial import TLKIMStage
 
         return TLKIMStage(microscope_name, device_connection, configuration, id)
-
+    elif device_type == "KINESIS":
```
AdvancedImagingUTSW, line 712: This device should also be added to the configuration wizard.

src/navigate/model/devices/camera/photometrics.py
```diff
@@ -66,8 +66,8 @@ def build_photometrics_connection(camera_connection):
     """
     try:
         pvc.init_pvcam()
-        # camera_names = Camera.get_available_camera_names()
-        camera_to_open = Camera.select_camera(camera_connection)
+        camera_names = Camera.get_available_camera_names()
+        camera_to_open = Camera.select_camera(camera_names[0])
```
AdvancedImagingUTSW, line 70: Does this assume only one camera?

src/navigate/model/devices/daq/ni.py
```diff
@@ -172,12 +172,14 @@ def set_external_trigger(self, external_trigger=None) -> None:
                 self.analog_output_tasks[
                     board_name
                 ].triggers.start_trigger.cfg_dig_edge_start_trig(trigger_source)
-                try:
```
AdvancedImagingUTSW, line 175: If it was throwing an error, but handled with a try/except statement, why did you have to remove it?

src/navigate/model/features/common_features.py
```diff
@@ -988,10 +988,10 @@ def pre_signal_func(self):
         self.z_stack_distance = abs(
             self.start_z_position - float(microscope_state["end_position"])
         )
-
+        # NOTE: added a quick fix for the focus stage moving in a negative direction during z-stack acquisitions. Somehow a negative step needs to be allowed. Flipping the axes direction did not have the desired outcome.
         self.start_focus = float(microscope_state["start_focus"])
```
AdvancedImagingUTSW, line 996: When you set the set z/f start position and z/f end position, it can't automatically sweep in the negative direction? If not, that seems like we should fix the logic immediately after those event calls.

PR #1069, opened 2025-03-01
add Ximea camera class file

annie-xd-wang:
This new device class file has been tested with software trigger and is working as expected. However, the device requires setting binning before configuring the ROI, which is the opposite sequence of our current workflow. We need to verify whether this change is compatible with other camera devices.

Additionally, the valid ROI values are somewhat tricky. It may be necessary to update the camera settings controller to improve user-friendliness.

We may need to have a discussion before merging this into our codebase.


AdvancedImagingUTSW:
Awesome. Thank you so much Annie.

AdvancedImagingUTSW (approved):
@annie-xd-wang - Another case where we will need to update the documentation. We should now provide information on the ximea camera...

PR #1073, opened 2025-03-19
Fix asi typo errors

annie-xd-wang:


AdvancedImagingUTSW (approved):
@annie-xd-wang - We should go through our documentation in the near-future and make sure that everything is up to date. I am concerned that our documentation is out of date with the new naming system. I also think we can essentially move forward with only the configurator now that we have tested it more substantially on multiple systems.

PR #1087, opened 2025-04-10
Abstract z stack acquisition

### LINKED ISSUE #1081: Abstract Z-Stack Acquisition Format
Currently we have the odd behavior where we can have a tiling axis and a stacking axis which share the same physical direction. Also, we have other microscopes that require multiple stages to move simultaneously during the acquisition.

For example, a laser scanning light-sheet microscope will scan the position of the illumination beam with a galvo and chase the laser beam by adjusting the position of the detection objective with a piezo.

For this case, it would be nice if we could specify multiple z-axes, perhaps as z0 and z1... in the configuration file. In the GUI one could specify which device you want to acquire the z-stack with:

Image

However, for the example I just stated, we would also have an option where you can scan multiple simultaneously. They would be going the same distance, but we would need to be able to control the offset between them. 

Galvo may go from 0 - 50 microns, but the piezo might go from 10 - 60 microns. The total amplitude of the scan would be the same, but we would need to be able to adjust the relative spacing of the two. 

Would need to specify which one is the primary z axis, perhaps defined as z0, and then the secondary z-axis would move with it if the user changes the z position via the stage control tab.

Primary/Secondary axes.

annie-xd-wang:


conorhughmcfadden (commented):
Works as intended on my machine!

A couple of things that may be good to add/consider:

1. The offset seems to be an absolute position in stage coordinates. Is this what we want? It seems it may be better to have the offset measured relative to the current position, e.g) relative stacking. If not, maybe a checkbox to give the option to have Offset tethered to the current position. Might be relevant in tiling when it may change pos to pos. If I misunderstand then disregard.

2. Would it be good to have a "Ratio" float input to scale the secondary stacking device relative to the primary? We could put it right next to Offset and allow it to be negative, for cases like mine where the stage direction is inverted.

AdvancedImagingUTSW:
Quick thoughts. Thank you for testing it @conorhughmcfadden.

_The offset seems to be an absolute position in stage coordinates. Is this what we want? It seems it may be better to have the offset measured relative to the current position, e.g) relative stacking. If not, maybe a checkbox to give the option to have Offset tethered to the current position. Might be relevant in tiling when it may change pos to pos. If I misunderstand then disregard._

I also thought that this would be a relative offset. The primary stage would decide the scan amplitude, which would be mirrored by any secondary stages. Say, the primary stage has a scan range of 20 - 80 microns, then the amplitude of the secondary stage would be 60 microns too. If the offset was set to -10 microns, then the secondary stage would scan from 10 - 70 microns. This logic only works if it is a stacking stage, and not a tiling stage. We should all make sure we are on the same page before implementing any changes. 

_Would it be good to have a "Ratio" float input to scale the secondary stacking device relative to the primary? We could put it right next to Offset and allow it to be negative, for cases like mine where the stage direction is inverted._

I think this gets back to us properly defining the stage axes. If the vector is inverted, then I think we would define this in the configuration file with a flip flag.

src/navigate/model/features/common_features.py
```diff
@@ -1163,11 +1161,14 @@ def signal_func(self):
                 self.model.pause_data_thread()
                 logger.info("Data thread paused.")
 
+            stack_pos = [
+                (f"{self.primary_z_axis}_abs", self.current_z_position),
+                (f"{self.primary_f_axis}_abs", self.current_focus_position)
+            ]
+            for axis, offset in self.secondary_stack_settings.items():
```
AdvancedImagingUTSW, line 1277: Previously, we only moved the Z and F.  Calculates the Z and F position each time...

Now we have more stages... If we want to move the stages with a variable definition of the offset, or with some coefficient (e.g., negative 1), we do so here.

AdvancedImagingUTSW: it is currently an offset based upon the primary z axis.

AdvancedImagingUTSW (approved):

PR #1100, opened 2025-06-05
ASI Galvo Integration

Daniel-Buckelew:
Bringing in Galvo changes to parent Navigate. This also includes some of the previous RFVC changes, due to how these are being packaged. Hopefully this won't cause any issues either.

PR #1107, opened 2025-06-30
ASI Daq integration, commented all ASI code

joelaugustine099:


AdvancedImagingUTSW:
@joelaugustine099 and I looked over the code and made some TODO comments. We compiled the documentation, and it looked correct. We tested the code on the bench top mesospim, and although we weren't testing the equipment explicitly, the software functioned as we hoped. Thus, no changes interfere with other microscope function.

Joel will test on the oscilloscope and report back.

The comments in PR #1100 were addressed here. This PR includes everything in #1100 and more, and should hopefully be the last of it.

src/navigate/model/devices/daq/asi.py
```diff
@@ -0,0 +1,271 @@
...
+        For this to work, all analog output and counter tasks have to be primed so that
+        they are waiting for the trigger signal.
+        """
+        # if self.is_updating_analog_task:
+        #     self.wait_to_run_lock.acquire()
+        #     self.wait_to_run_lock.release()
+
+        # turn on PLC cell 1 (Master Trigger)
+        try:
+            self.daq.logic_cell_on("1")
```
annie-xd-wang, line 251: It seems like this function doesn't currently use the full sweep time. It should either block for the full duration or signal when the acquisition task is finished.

joelaugustine099: I see the issue. This is functionality the UTD team did not implement and can be addressed in another PR.

PR #1108, opened 2025-07-01
Resizable window

### LINKED ISSUE #1106: Sidebar layout overflow in ‘Slices’ display mode
**Describe the bug**
When switching from "Live" to "Slices" in the Camera view, the right-hand sidebar (which contains LUT and display options) begins to stretch vertically, about one step per second. This quickly ruins the layout.

**To Reproduce**
1. Start Navigate and aquire some images (I ran the synthetic hardware version with "navigate -sh", used continuous scan, pressed "Aquire" and then "Stop" shortly after).
2. In the “Image Display” dropdown, change from Live to Slices.
3. Observe the sidebar on the right-hand side (LUT, Autoscale, etc.).
4. The UI expands vertically over time.

**Expected behavior**
Sidebar layout should remain fixed in height regardless of the display mode.

**Traceback**
No relevant traceback appeared when the bug occurred.

**Logs & Configuration Files**
Attached: Log files 2025-06-26-1138.zip from a session run with synthetic hardware (Navigate 0.0.8). 
Attached: Configuration files config.zip. Reflects the settings at the time the issue occurred, though older sessions seem to be overwritten by default. 

[config.zip](https://github.com/user-attachments/files/20923734/config.zip)

[2025-06-26-1138.zip](https://github.com/user-attachments/files/20923748/2025-06-26-1138.zip)

annie-xd-wang:
Fix issue #1106

### DIFF
--- a/src/navigate/controller/sub_controllers/camera_view.py
+++ b/src/navigate/controller/sub_controllers/camera_view.py
@@ -978,7 +978,7 @@ class BaseViewController(GUIController, ABaseViewController):
             width = max(
                 w_width - left_width - 16, 560 + self.view.lut.winfo_width()
             )
-            height = max(w_height - top_height - 38, 670)
+            height = max(w_height - top_height - 50, 670)
         else:
             width = event.width
             height = event.height - 24
@@ -1010,6 +1010,9 @@ class BaseViewController(GUIController, ABaseViewController):
             if widget != self.view.canvas:
                 if self.view.is_docked or widget.winfo_ismapped():
                     widget_height += widget.winfo_height() + 5
+                    if widget.winfo_height() < 30:
+                        widget_height += 30
+                    
         self.canvas_height = (
             height - widget_height - (50 if self.view.is_docked else -5)
         )
@@ -1113,8 +1116,7 @@ class CameraViewController(BaseViewController):
         # Slider Binding
         self.view.slider.bind("<Motion>", self.slider_update)
 
-        if platform.system() == "Windows":
-            self.resize_event_id = self.view.bind("<Configure>", self.resize)
+        self.resize_event_id = self.view.bind("<Configure>", self.resize)
 
         #: str: The display state.
         self.display_state = "Live"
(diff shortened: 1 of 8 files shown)


src/navigate/controller/sub_controllers/camera_view.py
```diff
@@ -964,15 +964,28 @@ def resize(self, event: tk.Event) -> None:
...
+            w_width = self.parent_controller.view.winfo_width()
+            w_height = self.parent_controller.view.winfo_height()
+            width = max(
+                w_width - left_width - 16, 560 + self.view.lut.winfo_width()
+            )
+            height = max(w_height - top_height - 50, 670)
+        else:
+            width = event.width
+            height = event.height - 24
+
```
AdvancedImagingUTSW, line 985: The widths and heights make great sense. But what about these other numbers (16, 560, 50, 670, 24)? Would it make sense to have these not hard-coded?

src/navigate/controller/sub_controllers/camera_view.py
```diff
@@ -984,17 +997,29 @@ def refresh(self, width: int, height: int) -> None:
         height : int
             Height of the window.
         """
```
AdvancedImagingUTSW, line 999: Same with these hard-coded values.

AdvancedImagingUTSW (approved):
Works great on my computer. I also tagged @joelaugustine099 to review this, just to expose him to a different part of the codebase and to introduce him to the process.

Joel, switch to this PR branch and try adjusting the size with and without the tabs popped out. Make sure that it is stable. Also, look over the code, and make sure that it is logically clear (as much as possible).

src/navigate/controller/sub_controllers/camera_view.py
annie-xd-wang: Yeah, I'm not a fan of those hard-coded numbers either. Ideally, we’d avoid them — but there are a couple of practical reasons they’re in place for now:

1. When the window is rendered as a tab vs. a popup, the layout behaves slightly differently, so we adjust the height a bit to stabilize the rendering.

2. Widget sizes differ between Windows and macOS, and these offsets help prevent the window from constantly resizing itself due to small layout shifts.

I can look into it further to see if we can remove those numbers.

joelaugustine099:
It looks good. Code is clear. The rendering is a bit choppy on my computer but I think that's what is intended (so that it isn't re-rendering every time the window is resized). 

One thing that could be avoided is that when I extend the window, the features flash in their old position before appearing in the new position. This is just aesthetics though - it functions great.

joelaugustine099 (approved):

PR #1110, opened 2025-07-02
Provide Additional User Context

AdvancedImagingUTSW:
Now it should be more obvious when the MIP display is not enabled, who to use it.

You should be able to right-click on both the histogram and the MIP display and enable/disable it. MIP display only works in the z-stack acquisition mode.


joelaugustine099 (approved):
Everything looks good. MIP display and histogram can be enabled/disabled.

PR #1117, opened 2025-07-29
Fix: undefined variable in KST101Stage.connect() (serialnum → serial_…

juhelh:
## Summary

Fixes a `NameError` in the Thorlabs KST101 stage controller class (`KST101Stage`) where the variable `serialnum` was used but never defined.

### Changes

- Replaced all uses of `serialnum` with the correct argument name `serial_number`.
- Renamed `available_serialnum`  to `available_serial_numbers` for clarity.

### Impact

Prevents a runtime crash when trying to connect to Thorlabs stepper stages using this driver.

### DIFF
--- a/src/navigate/model/devices/stage/thorlabs.py
+++ b/src/navigate/model/devices/stage/thorlabs.py
@@ -341,14 +341,14 @@ class KST101Stage(StageBase):
 
         # Open the same serial number device if there are several devices connected to the
         # computer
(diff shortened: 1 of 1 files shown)


AdvancedImagingUTSW:
Thank you. Good catch. We haven't used this controller internally since we refactored how devices are loaded.

AdvancedImagingUTSW (approved):

PR #1118, opened 2025-07-29
Fix: update internal position after ASI filter and slider filter wheel move

juhelh:
- Updates internal position attributes (self.filter_wheel_position, self.dichroic_position) after successful hardware move.

- Wraps hardware move calls in try/except with logging and exception re-raise.

- Catches a failed move early. Makes sure internal position always follows the actual position.


juhelh:
Accidently pushed the wrong code (sorry!). Will write a new PR that shortly that replaces this one.

PR #1120, opened 2025-07-29
Error propagation

### LINKED ISSUE #1121: Custom Exception Class Triggers Tk Messagebox
As discussed in #1119, we could create a special class of exceptions. We could initialize these at the base layer for each device class, or even possibly every base class could inherit an exception type which bubbles up to the model and then directly informs the user of the problem. Each exception would require an informative message prompt.

We do not want this called all the time, such as when we have the NI DAQ exceptions that we automatically handle. But rather, when a preventable configuration-based error occurs.

annie-xd-wang:
This PR introduces the ability to propagate error messages to the main thread and display them to users in a popup window.


AdvancedImagingUTSW:
Ready for review?

annie-xd-wang:
It's ready for review. Any exception that should be propagated to the user must inherit from UserVisibleException and be raised explicitly.

AdvancedImagingUTSW (approved):

PR #1122, opened 2025-07-31
Add Tecan XCalibur pump interface and test suite

juhelh:
This PR introduces a new device interface for the Tecan XCalibur syringe pump, including a corresponding test suite.

Additions:
- `TecanXCaliburPump` class under `devices/pump/`, implementing core pump functionality.
- `test_tecan_x_calibur.py` under `tests/pump/`, using a custom `FakeSerial` class to simulate command/response behavior.

Note on integration:
Integration into configuration.yaml and Navigate’s startup flow is left to the core team, as it may depend on internal conventions. I’m happy to assist though in a follow-up PR if helpful.

Note on try/except usage:
In another PR we discussed that try/except blocks may not always be preferred in Navigate. I'm now aware of that design philosophy, but I chose to use it here because I believe it can offer more flexibility.

Specifically, if these methods are called higher up in the stack, developers can choose either to:
- Not wrap them in try/except, allowing exceptions to propagate and halt execution (e.g. to stop an experiment), or
- Wrap them and handle errors more gracefully (e.g. with `except pass` or logging + retry).

I imagine both cases have their place, though I’ll admit I’m no expert in microscopy control. I'm absolutely open to changing this if it doesn’t align with Navigate’s philosophy. Just let me know!

### DIFF
--- a/src/navigate/model/devices/pump/tecan.py
+++ b/src/navigate/model/devices/pump/tecan.py
@@ -1,0 +1,409 @@
+# Copyright (c) 2021-2025  The University of Texas Southwestern Medical Center.
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted for academic and research use only (subject to the
+# limitations in the disclaimer below) provided that the following conditions are met:
+
+#      * Redistributions of source code must retain the above copyright notice,
+#      this list of conditions and the following disclaimer.
+
+#      * Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer in the
+#      documentation and/or other materials provided with the distribution.
+
+#      * Neither the name of the copyright holders nor the names of its
+#      contributors may be used to endorse or promote products derived from this
+#      software without specific prior written permission.
+
+# NO EXPRESS OR IMPLIED LICENSES TO ANY PARTY'S PATENT RIGHTS ARE GRANTED BY
+# THIS LICENSE. THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+# CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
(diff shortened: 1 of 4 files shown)


annie-xd-wang:
The device code is well-structured and works independently as a device API. However, if we integrate this file directly as the device layer (rather than using it as the lower-level API), several changes are needed.

1. File naming – The Python file should be named after the manufacturer. For this device, it should be:

`tecan.py`

2. Class naming – The class name should match the device model name. In this case:

`class XCaliburPump:`

3. Device layer interface – Each device class in our device layer must follow this constructor signature: 

`def __init__(
    self,
    microscope_name: str,
    device_connection: Any,
    configuration: Dict[str, Any],
    device_id: int = 0,
):`
4. Serial device connection – For serial-based devices, the following class method is required:

`@classmethod
def connect(cls, port, baudrate=115200, timeout=0.25):`

These changes will make the implementation consistent with our device layer structure and naming conventions.

AdvancedImagingUTSW:
@juhelh - If you can make the changes that @annie-xd-wang recommended, I will merge it into develop.  Regarding the try/except usage, we should be able to raise a `UserVisibleException`, which will provide the user with a pop-up, rather than crashing the software. 

If you would like to meet and discuss the best way to further integrate your hardware, please let me know.

juhelh:
Hi @annie-xd-wang and @AdvancedImagingUTSW! Sorry for a rather slow response. The recommendations from @annie-xd-wang should be fairly quick to implement, I will have a look at it and get back to you. 

About the exception handling: If you want I can try to change all exceptions to UserVisibleException? In priniciple I think I prefer the idea of using standard try/except at low-level code like this, and making decisions about how to catch the errors (e.g. doing nothing, stopping or raising UserVisibleException) further up in the code, but I also understand that this might require more changes, and can be a bit cumbersome. So I'm totally open to using UserVisibleException as a more practical solution. 

Will start working on it. I'll let you know if I have more questions along the way.

juhelh:
Hi! 

I accidentally force-pushed more changes than intended (some unrelated local work got included). I'm cleaning it up now. Will push a clean branch shortly with just the intended Tecan pump refactor.

Apologies! I'm new to working with git (as you may be able to tell...).

juhelh:
Alright, I've now created a new PR with the updates! Only uploaded the correct files this time. Please check out that code instead, and thank you for your patience - I will hopefully learn from my mistakes and at least not make this silly github mistake again.

PR #1127, opened 2025-08-15
Add Daheng camera class and tests

juhelh:
## Summary

- **New `DahengCamera` class in `daheng.py`**  
  Implements a camera interface class using the SDK `gxipy`. It's designed to follow Navigate's architecture.

- **Unit test file `test_daheng.py`**  
  Includes tests for class behavior. Mocks the critical SDK interfaces such as `DeviceManager`, `FeatureControl`, and `DataStream`.

- **Exception and logging policy followed**  
  Uses `UserVisibleException` for GUI error messages, and `logging.warning()` for non-fatal issues.

- **Handles unsupported features with stubs**  
  Some methods such as `get_readout_time`, `get_line_interval`, and `set_light_sheet_exposure_time` are included as stubs. This is simply because the MER2-1220-32U3C model does not have the functionality needed, but we don't want anything to break if camera.get_line_interval() e.g. is called somewhere else in the code.

---

## gxipy SDK Installation

The Daheng SDK (`gxipy`) is **not** available on PyPI. To use this camera class:

1. Download the SDK from [https://www.daheng-imaging.com/](https://www.daheng-imaging.com/)
2. Locate the `gxipy` folder inside the SDK (typically under `Development/Samples/Python/`)
3. Install it manually with `pip install /path/to/gxipy`.

An `ImportError` message has been added to the top of the file to guide users who try to use the class without having installed the SDK.  
Maybe it would also be useful to add a note in the README about this manual installation step for new users?


AdvancedImagingUTSW:
@juhelh - Sorry for the delay. This is one of the first times that someone has integrated such a core device, and I think we were figuring out how best to proceed given that we do not have the device internally. Is it working for you? We may have a few things to ask that you add before merging... 

Thanks,
Kevin

annie-xd-wang (dismissed):

PR #1136, opened 2025-09-12
Update shear parameters

tngo2019:
Tested new affine shear and tiling, it worked as intended.


AdvancedImagingUTSW (approved):
We should also document the shearing and rotation BDV parameters in the documentation.

Issue #1142, opened 2025-09-17
LUDLFilterWheel no longer works with new serial device monitoring updates


conorhughmcfadden:
**Describe the bug**
If I have a LUDLFilterWheel as the filter wheel in my config file, I cannot get an image when and only get frame timeouts. It works fine on my previous branch `ao-optimizers-test` which is unmerged, but from 2 months ago. Most likely the error has something to do with the recent performance monitoring updates, probably LUDLFilterWheel was not updated like the rest of the devices, because no one else really uses LUDL stuff.

***NOTE:*** Setting my filter wheel to be `synthetic` fixes things, so it's definitely in LUDLFilterWheel. 

**To Reproduce**
1. Start Navigate
2. Try to do a Single Acquisition
3. Wait and get a frame timeout
4. On Exit, we get an error.

**Error Message on Console after Exit**
```console
frame timeout. Verify timeout exceeds exposure time. If applicable, check external trigger source.
Exception ignored in: <function LUDLFilterWheel.__del__ at 0x0000018705093B80>
Traceback (most recent call last):
  File "C:\Users\vastopmv3\Documents\GitHub\navigate\src\navigate\model\devices\filter_wheel\ludl.py", line 170, in __del__
    self.close()
  File "C:\Users\vastopmv3\Documents\GitHub\navigate\src\navigate\model\devices\filter_wheel\ludl.py", line 164, in close
    self.set_filter(list(self.filter_dictionary.keys())[0])
  File "C:\Users\vastopmv3\Documents\GitHub\navigate\src\navigate\model\devices\filter_wheel\ludl.py", line 153, in set_filter
    self.sio.flush()
  File "C:\Users\vastopmv3\Documents\GitHub\navigate\src\navigate\tools\decorators.py", line 100, in wrap_func
    "args": display_args(*args) if display_args else "Hidden",
  File "C:\Users\vastopmv3\Documents\GitHub\navigate\src\navigate\model\device_startup_functions.py", line 187, in <lambda>
    display_args=lambda d: f"{str(port)}-{d.decode(errors='ignore')}"
AttributeError: 'memoryview' object has no attribute 'decode'
Exception ignored in: <module 'threading' from 'C:\\Users\\vastopmv3\\.conda\\envs\\navigate\\lib\\threading.py'>
SystemExit:
Exception in thread Thread-1:
BrokenPipeError: [WinError 109] The pipe has been ended

During handling of the above exception, another exception occurred:

EOFError
```

[...]

PR #1143, opened 2025-09-19
fix: decode bytes and memoryview safely

annie-xd-wang:
Fix issue #1142


conorhughmcfadden:
I tested it with my LUDLFilterWheel included and it now appears to be working! I can acquire images and the filter wheel switches correctly when I change channels.

AdvancedImagingUTSW (approved):

PR #1151, opened 2025-11-07
Cache zstack stage position

annie-xd-wang:


AdvancedImagingUTSW (approved):

AdvancedImagingUTSW:
Upon inspection, we realized that the position of the stages were being queried during a z-stack, adding some latency. This PR makes sure that the cached positions are retrieved, and that the actual devices are not communicated with. It also adds the tests to make sure that this stays the case.

PR #1157, opened 2025-12-03
Fix ni filter wheel toggle issue

annie-xd-wang:
NI-DAQ–controlled filter wheels rotate/toggle on every 5 V edge trigger. If the wheel is already at the requested position, sending another trigger will shift it to the wrong position. This PR caches the current position and suppresses triggers when the wheel is already at the target slot.

### DIFF
--- a/src/navigate/model/devices/filter_wheel/ni.py
+++ b/src/navigate/model/devices/filter_wheel/ni.py
@@ -55,7 +55,7 @@ logger = logging.getLogger(p)
 class NIFilterWheel(FilterWheelBase, NIDevice):
     """DAQFilterWheel - Class for controlling filter wheels with a DAQ."""
 
-    filter_wheel_value = None
+    filter_wheel_value = {}
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


AdvancedImagingUTSW:
`filter_wheel_value` is implemented as a class variable instead of an instance variable, which will cause issues if multiple filter wheels are in use. Thus, the current implementation uses type(self).filter_wheel_value to read/write this value. This will...
- All NIFilterWheel instances share the same filter_wheel_value
- If multiple filter wheels exist, they cannot track their positions independently
- One filter wheel's position change will overwrite another's cached value

is this intentional @annie-xd-wang? Or should it be an instance variable?

```
def __init__(self, microscope_name, device_connection, configuration, device_id=0):
    super().__init__(microscope_name, device_connection, configuration, device_id)
    self.wait_until_done_delay = self.device_config["filter_wheel_delay"]
    self.filter_wheel_task = None
    self.filter_wheel_value = None 
```

AdvancedImagingUTSW (approved):

PR #1158, opened 2025-12-05
Remove unnecessary sleep and add performance TODO

AdvancedImagingUTSW:
Removed a redundant time.sleep call in SyntheticStage's move_absolute method to improve performance. Added a TODO comment in Microscope to highlight the cost of repeated dictionary lookups for image_mode during z-stack acquisition, suggesting future optimization.


joelaugustine099 (dismissed):
Looks good

PR #1163, opened 2026-01-07
Fix zoom image extraction for transposed views

AdvancedImagingUTSW:
Adds conditional logic to handle image extraction when the transpose flag is set, ensuring correct slicing of image data for both transposed and non-transposed views. Also refactors grid placement for SNR palette for consistency.


AdvancedImagingUTSW:
@annie-xd-wang - Tai has confirmed that the change fixed the problem, which was that when the image was rotated in the GUI, the logic for the zoom did not take that into account. Now the zoom works regardless of if the image is transposed.

PR #1166, opened 2026-01-14
1101 feature request unified global coordinate system across all microscopes

### LINKED ISSUE #1101: Feature Request: Unified Global Coordinate System Across All Microscopes
### Feature Request: Unified Global Coordinate System Across All Navigate Microscopes

---

#### **Problem / Motivation**
Smart-acquisition routines assume every microscope shares the same axis conventions, but each rig currently has different stage polarities and camera orientations.

*Symptoms*
- **Z-axis confusion:** Positive *Z* moves **toward** the objective on some scopes and **away** on others.
- **Camera inversions:** Image X/Y directions vary between systems.
- **Multi-scale mis-registration:** Coordinate transforms aren’t consistent across resolutions.
- **Crash risk:** A user/workflow can drive the stage into an objective if driven the wrong direction.

A single, explicit coordinate system—plus per-microscope flip-flags—will eliminate hidden assumptions, simplify code, and make automated routines safe and portable.

---

#### **Proposed Solution**

1. **Adopt an Absolute Stage-Space Convention**

   | Axis  | **+ Direction (physical)**               | Crash-Risk Notes                     |
   |-------|------------------------------------------|--------------------------------------|
   | **Z** | Toward detection objective               | Always the “crash” direction         |
   | **Y** | Toward illumination objective            | ­                                    |
   | **X** | Away from optical-table front edge       | ­                                    |
   | **F** | Same sense as **+Z** (focus motor)       | ­                                    |
   | **θ** | Counter-clockwise (viewed from above)    | ­                                    |

For multi-view illumination systems, the primary microscope would define the coordinate system. Additional working definitions are needed for the upright ASLM and OPM rigs where the specimen moves obliquely relative to the classical axes. 

2. **Define Camera-Space Rules**

   * +Y stage ⇒ image moves **up** (screen coordinates).  
   * +X stage ⇒ image moves **right**.  
   * Per-camera `camera_flip_flags` correct any mismatch.

3. **Config-Level Mapping**

   ```yaml
   stage_flip_flags:
     flip_x: false
     flip_y: true
     flip_z: false
     flip_theta: true

   camera_flip_flags:
     flip_x: true
     flip_y: false

Provide GUI helper:

Create a popup window that allows someone to readily change the flip flags without having to reboot the software. 

4.	Stitching Workflows
•	Needs to be extensively tested with BigDataViewer to guarantee that we do not disrupt downstream workflows.
 
5.	Safety & Regression Tests	
•	Unit test: Simulate flip matrices; verify move_relative(+z) maps to a positive internal delta toward the crash direction.
•	Integration test: Jog +10 µm in each axis and assert expected image displacement on every microscope. This would also be a good way to calibrate the stages when they move in the lateral dimensions. 

6.	Documentation	•	Add “Coordinate System” page with diagrams, rationale (link to Slack discussion), and migration guide.

Priority: P0 – prerequisite for reliable multi-microscope smart routines

AdvancedImagingUTSW:
@annie-xd-wang - I need to test this and make sure that the new popup works for setting the camera flip flags from within the software. Once that is confirmed to be functioning, we can harmonize the microscopes to have consistent movement with the camera and stages...  And then finally, reduce the number of permutations that we perform with the image...

### DIFF
--- a/docs/source/02_user_guide/04_microscope_setup/01_software_configuration/coordinate_system.rst
+++ b/docs/source/02_user_guide/04_microscope_setup/01_software_configuration/coordinate_system.rst
@@ -1,0 +1,207 @@
+.. _coordinate_system:
+
+============================
+Microscope Coordinate System
+============================
+
+This section details the expected coordinate system conventions for **navigate**-controlled microscopes.
+
+-----------------
+
+Problem / Motivation
+====================
+
+Smart-acquisition routines assume every microscope shares the same axis conventions, but each rig currently has different stage polarities and camera orientations. To ensure consistent behavior across different microscope setups, **navigate** defines a standardized coordinate system that should be followed when configuring new microscopes.
+
+By adhering to these conventions and properly configuring the flip flags in your ``configuration.yaml`` file, you can ensure that:
+
+* Stage movements are intuitive and predictable
+* Camera image orientation matches stage movements
+* Multi-microscope configurations work correctly
+* Smart-acquisition features function as expected
+
+-----------------
+
+Standard Light-Sheet Microscope
+================================
+
+For the standard light-sheet microscope, with a single illumination and detection objective, we define the following coordinate system:
+
+Axis Definitions
+----------------
+
+The physical coordinate system is defined relative to the microscope geometry:
+
+**Z-Axis (Detection Axis)**
+  * **-z** is towards the detection objective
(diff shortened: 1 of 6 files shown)


annie-xd-wang:
Confirmed: the popup window works correctly for setting the flip flags within the software.

PR #1169, opened 2026-01-22
Refactor frame rate update to use model event

AdvancedImagingUTSW:
Moved frame rate calculation and update logic from the controller to the model, which now sends accurate frame rate updates via the event queue. The controller receives these updates and refreshes the GUI accordingly. Updated related test to simulate the new frame rate event mechanism.


joelaugustine099 (approved):
Looks good

PR #1170, opened 2026-01-24
Fix crosshair error when flipping non-square images

annie-xd-wang:


AdvancedImagingUTSW (approved):
Made one small change so that the crosshair coordinates are also transposed, rather than reset to the center coordinate.

PR #1172, opened 2026-02-03
Add stage position caching and safe config access

### LINKED ISSUE #1159: Avoid Repetitive Proxy Dict Calls
**Describe the bug**
Calling the proxy dictionary costs ~3 ms, and is performed for every step in the z-stack. To optimize performance, we should retrieve the image mode once and avoid repeated dictionary lookups.

See `navigate/model/microscope.py`:

```
        if self.configuration["experiment"]["MicroscopeState"]["image_mode"] in (
            "z-stack",
            "customized",
        ):
            # cache stage positions in z-stack and customized modes.
            self.ask_stage_for_position = False
            for axis_key in pos_dict.keys():
                axis = axis_key[: axis_key.index("_")]
                self.ret_pos_dict[f"{axis}_pos"] = pos_dict[axis_key]
        else:
            self.ask_stage_for_position = True
```

Eliminating this delay would improve the frame rate ~5 Hz.

AdvancedImagingUTSW:
Use safe dict .get lookups in GUI controllers to avoid KeyError when nested settings are missing. Introduce Microscope.cache_stage_positions and set_stage_position_cache_policy(image_mode) to enable caching of stage positions for z-stack/customized modes, reducing repeated config lookups and per-step overhead. Wire policy application in Model during initialization and update tests to call the policy setter and assert caching behavior. Also add AGENTS.md with project notes and developer workflows.

### DIFF
--- a/.github/AGENTS.md
+++ b/.github/AGENTS.md
@@ -1,0 +1,54 @@
+---
+name: navigate-agent
+description: You are a very careful software engineer and an expert in fluorescence microscopy, hardware communication, and you prioritize high-performance code.
+---
+
+## Project knowledge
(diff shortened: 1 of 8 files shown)


AdvancedImagingUTSW:
@conorhughmcfadden - Could you test this branch and see if you can now achieve something closer to 100 Hz?

annie-xd-wang (approved):

PR #1173, opened 2026-02-03
Drop queued frames in live display

AdvancedImagingUTSW:
When in live mode, drain any queued frames from show_img_pipe so only the most recent image is processed, preventing visible display lag from a pipe backlog. Count dropped frames, log a debug message when frames are dropped, and include dropped frames in images_received. Preserve existing handling for the "stop" sentinel and non-int image ids.

### DIFF
--- a/src/navigate/controller/controller.py
+++ b/src/navigate/controller/controller.py
@@ -1217,7 +1217,22 @@ class Controller:
                 break
             # Receive the Image and log it.
             image_id = self.show_img_pipe.recv()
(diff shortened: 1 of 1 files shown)


AdvancedImagingUTSW:
@annie-xd-wang - We now include another safeguard which should prevent a laggy display. @conorhughmcfadden has tested it, and it appears to be functional. Camera display no longer lags significantly.

annie-xd-wang (approved):

PR #1175, opened 2026-02-04
Rotation tiling

AdvancedImagingUTSW:
This change fixes BDV metadata so per‑view stage rotation (theta) is encoded as a true rotation about the volume center instead of the first voxel. The BDV affine now explicitly applies the required translation offset (t' = t + c − R·c), ensuring rotations are centered on the volume in voxel coordinates. Tile IDs are still derived from translation only, so angle changes don’t create new tiles. A small unit test was added to validate tile/angle mapping in the generated XML.


AdvancedImagingUTSW:
@annie-xd-wang - This now includes some important updates to the multiposition table. I noticed that occasionally it would enter a resize death loop. A resize event would trigger more resize events, and the software would freeze. This now includes a safeguard to prevent that. Otherwise, the x and y pixel positions are saved in the yaml file, but not included in the table. I would like to make sure it works with additional stage axes that are not the normal x, y, z, f, theta.

pytest.ini
```diff
@@ -2,8 +2,12 @@
 addopts =
     --strict-markers
     -m "not hardware"
-    --cov=./
```
AdvancedImagingUTSW, line 5: This limits the code coverage to the src directory. Implemented to reduce noise in the CI routine.

src/navigate/controller/sub_controllers/multiposition.py
```diff
@@ -111,6 +113,64 @@ def _refresh_table_view(self) -> None:
...
+        """Normalize known column aliases and case."""
+        normalized_df = df.copy()
+        normalized_df.columns = [str(col).upper() for col in normalized_df.columns]
+        stage_axes = self.parent_controller.configuration_controller.stage_axes
+        if "theta" in stage_axes:
+            if "THETA" not in normalized_df.columns and "R" in normalized_df.columns:
+                normalized_df = normalized_df.rename(columns={"R": "THETA"})
+            elif "THETA" in normalized_df.columns and "R" in normalized_df.columns:
+                normalized_df = normalized_df.drop(columns=["R"])
+        return normalized_df
```
AdvancedImagingUTSW, line 133: Avoids problems associated with Theta being named R

AdvancedImagingUTSW:
One other comment. How the rotation is implemented in the BDV still needs to be verified. But, in theory, I think it looks approximately right. Will need to confirm with a sample. 
image

Also, quickly pushing a quick commit to avoid excessive printing of a matplotlib warning...

annie-xd-wang (approved):

PR #1193, opened 2026-02-27
Pause/resume on theta move; wait for stage target

AdvancedImagingUTSW:
Add theta_move_tolerance and stage move verification to avoid imaging while rotating. Implement _wait_for_stage_target with timeout and polling, and call it after stage moves. Use absolute deltas where appropriate and include theta movement in pause/resume decisions for MoveToNextPositionInMultiPositionTable, ZStackAcquisition and ASIZStackAcquisition. Ensure Microscope sets ask_stage_for_position before single-axis queries. Add unit tests (test_stage_move_pause_behavior.py) to cover theta-triggered pauses and large translations.

### DIFF
--- a/src/navigate/model/features/common_features.py
+++ b/src/navigate/model/features/common_features.py
@@ -755,6 +755,19 @@ class MoveToNextPositionInMultiPositionTable:
         #: int: The stage distance threshold for pausing the data thread.
         self.stage_distance_threshold = 1000
 
+        #: float: Any theta movement larger than this value triggers a data-thread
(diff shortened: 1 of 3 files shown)


AdvancedImagingUTSW:
This did not work, so the problem remains. Theta movement is very slow, and for some reason our polling of the Tiger Controller does not accurately wait until stage movement is complete. Rotation begins and is interrupted by the acquisition of another z-stack in the multi position table. Thereafter, the rotation stage seems blocked; no further stage move attempts work until the Tiger Controller is power cycled.

PR #1198, opened 2026-03-02
Support Piezo as stage

Johnhaug223:
fixes default_axes_sequence to include axes with type "p" (piezo) as a motor axis


AdvancedImagingUTSW (commented):
Make sure that this works on other ASI systems and let me know. One thing that pops up is that set_speed_as_percent_max now dereferences self.default_axes_sequence before validating it, which could raise TypeError instead of ASIException.

AdvancedImagingUTSW (approved):

PR #1222, opened 2026-05-16
Constrain stack ranges to stage limits

### LINKED ISSUE #969: Stage Limits 2.0
Currently people can type values beyond the limits of the stages. For example, with the 200 micron range piezo, people can go from 0 - 200 and everything is fine. If you type -100 - 100, the stack won't move for half the time. People get confused. 

Would be nice to have it set at the widget level as well. There is a _to and _from parameter in the widgets that can do this.

AdvancedImagingUTSW:
## Summary
- derive stack Z/F start and end widget bounds from active microscope stage limits in `configuration.yaml`
- keep `gui_configuration.yml` stack position bounds as the fallback when stage limits are disabled or unavailable
- refresh stack bounds after origin updates, selected stack-device changes, and stage-limit setting changes
- validate saved/programmatic stack bounds during acquisition verification

Fixes #969.

## Validation
- `conda run -n navigate python -m pytest test/controller/sub_controllers/test_channels_tab.py -q`
- `conda run -n navigate python -m pytest test/controller/test_controller.py::test_execute_stage_limits test/controller/test_controller.py::test_execute_update_stage_limits_refreshes_stack_position_limits -q`
- `conda run -n navigate python -m pytest test/controller/sub_controllers/test_channels_tab.py::test_update_z_steps test/controller/sub_controllers/test_channels_tab.py::test_update_start_position test/controller/sub_controllers/test_channels_tab.py::test_update_end_position -q`
- `ruff check src/navigate/controller/controller.py src/navigate/controller/sub_controllers/channels_tab.py test/controller/sub_controllers/test_channels_tab.py test/controller/test_controller.py`
- `git diff --check`

### DIFF
--- a/src/navigate/controller/controller.py
+++ b/src/navigate/controller/controller.py
@@ -1044,6 +1044,7 @@ class Controller:
             microscope_name = args[0]
             if microscope_name == self.configuration_controller.microscope_name:
                 self.stage_controller.initialize()
+                self.channels_tab_controller.update_stack_position_limits()
             self.threads_pool.createThread(
                 resourceName="model",
                 target=self.update_stage_limits,
@@ -1166,6 +1167,7 @@ class Controller:
 
         elif command == "stage_limits":
             self.stage_controller.stage_limits = args[0]
+            self.channels_tab_controller.update_stack_position_limits()
             self.threads_pool.createThread(
                 resourceName="model",
                 target=lambda: self.model.run_command("stage_limits", *args),
@@ -1434,7 +1436,7 @@ class Controller:
         None
         """
         e = RuntimeError
-        while e == RuntimeError:
+        while e is RuntimeError:
             try:
                 self.model.run_command("stop")
                 e = None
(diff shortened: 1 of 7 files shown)


AdvancedImagingUTSW:
Added context on the finite-acquisition drain added in `85fba93ce`:

The Windows CI failure in `test_single_acquisition` exposed a race between the signal side and the data side of finite acquisitions. For `single` mode, the signal loop can complete all selected channels and set `stop_acquisition = True` before the data loop has drained the final frame(s) already generated by the synthetic camera. Since `run_data_process()` used `while not self.stop_acquisition`, the data thread could exit immediately, send the `"stop"` sentinel, and leave the last generated frame unread. That is how CI ended up seeing 2 delivered frames for a 3-channel acquisition.

The new `signal_acquisition_complete` flag distinguishes natural finite-acquisition completion from a user/manual stop. When the signal side completes naturally, the data loop is allowed to continue only while the data container is still incomplete, so pending frames are drained before `"stop"` is sent. Live/manual stop behavior remains prompt because live mode and explicit stop requests do not enable this drain path.

annie-xd-wang (approved):

PR #1226, opened 2026-06-17
Optional "pulse" waveform option for galvo

conorhughmcfadden:
Resonant galvos work differently from linear galvos: they run at a fixed frequency with only the amplitude being controllable by a DC input voltage. I'm tired of fiddling with the little USB voltage generators every time I image. The goal is to have a simple pulse which can run the resonant galvo while acquiring, but is otherwise silent.

**Example config**
```
galvo:
  amplitude: 0.3
  hardware:
  channel: PCIe-6738-A/ao8
    max: 5
    min: 0
    name: daq
    type: NI
  name: resonant_galvo
  offset: 0
  phase: 0.0
  waveform: pulse
 ```

Pretty basic pulse, delayed by camera delay and lasting sweep time duration.
Creates a `waveforms.single_pulse` for the galvo with just amplitude as a user flex parameter.

### DIFF
--- a/docs/source/02_user_guide/01_supported_hardware/galvo.rst
+++ b/docs/source/02_user_guide/01_supported_hardware/galvo.rst
@@ -15,7 +15,12 @@ National Instruments
 Multiple types of galvanometers have been used, including Cambridge Technologies/Novanta, Thorlabs, and ScannerMAX. Each of these devices are externally controlled via analog signals delivered from an NI-based data acquisition card.
 
 For NI and synthetic galvos, the software supports the following waveform options:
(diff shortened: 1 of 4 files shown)


AdvancedImagingUTSW:
@conorhughmcfadden I pushed an update in `cf3bdf580` that exposes/documents the new `pulse` galvo waveform and fixes the pulse timing conversion.

The key correction is that `single_pulse()` expects `delay` and `pulse_width` as percentages of the sweep. The galvo pulse branch was using `camera_delay / sweep_time` as a fractional ratio, so it now converts that delay to percent before generating the waveform.

Could you please pull the latest branch and test the software again on your machine with your resonant galvo setup?

AdvancedImagingUTSW (approved):

PR #1227, opened 2026-07-02
Live 3D Volume Rendering with OpenGL

conorhughmcfadden:
## Live 3D Volume Rendering with OpenGL

This PR adds an optional real-time 3D volume renderer to navigate. It is gated behind an optional dependency group and is fully backwards-compatible — navigate is unaffected if OpenGL is not installed.

---

### System Requirements

- A GPU with OpenGL 3.3+ support (any modern NVIDIA, AMD, or Intel integrated GPU)
- Windows / macOS / Linux (tested on Windows 10/11 with NVIDIA hardware)

### Installation

```bash
# Development install with optional OpenGL dependencies
pip install -e .[opengl]
```

This pulls in: `PyOpenGL`, `PyOpenGL-accelerate`, `glfw`, `PyGLM`, `tkinterdnd2`.

---

#### New: OpenGL backend (`view/display_backends/gl_backend.py`)

The core renderer lives in `GLVolumeViewBackend`. To avoid blocking the Tkinter UI, all OpenGL work runs on a **dedicated render thread**:

- `start()` / `stop()` spin up/down a GLFW window and render loop on that thread.
- Python sends work to the GPU via two thread-safe queues: `cmd_q` for shader/state commands (lambdas executed on the GL thread) and `data_q` for raw slice uploads.
- The thread signals `is_ready` once the GL context is initialized, so `start()` blocks safely until rendering is available.

The renderer uses **raymarching**: a full-screen quad is rendered and each pixel fires a ray through the 3D volume texture. The ray is clipped to the volume bounding box, then marches forward in steps (`stepWorld`), sampling the 3D texture and accumulating colour with Beer-Lambert opacity blending. The loop exits early when the ray is fully opaque. Shear-angle correction is applied per step to support Oblique Plane Microscopy (OPM) datasets.

Up to 5 channels are supported simultaneously. Each channel has its own 3D texture and a row in a shared transfer-function texture (colour, min/max, gamma).

The `Camera` class (also in this file) handles orbit rotation, pan, dolly, and perspective/ortho toggle (Tab). Hold Shift during rotation to snap to 15° increments.

#### Integration into the main application (`controller/sub_controllers/camera_view.py`)

The backend is added to `BaseViewController` by **composition** — it is never imported at module level. Instead, it is imported lazily inside `_OpenGL_initialize_volume_rendering_backend()` which is called once during `initialize_non_live_display()`:

```python
try:
    from navigate.view.display_backends.gl_backend import GLVolumeViewBackend
    self.gl_volume_view_backend = GLVolumeViewBackend()
except ModuleNotFoundError:
    logger.info("Failed to initialize GLVolumeViewBackend...")
    self.view.after(100, self.view.volume_frame.grid_forget)
    return
```

If the import fails (OpenGL not installed), the Volume Rendering panel is hidden and navigate continues exactly as before.

When running, incoming z-slices are pushed to `data_q` during acquisition. The **Volume Rendering** panel in the display notebook exposes four controls:

| Control | Description |
|---|---|
| Enable | **Starts/stops the GLFW window** |
| Shear Angle | OPM correction angle (°) |
| Opacity | Global density multiplier |
| World Step | Ray march step length (larger = faster, coarser) |

#### Live Deskewing for OPM-like Sheared Stacks

OPM and other obliquely sectioned microscopy modalities acquire stacks in a sheared format, which requires offline deskewing. I have implemented [PetaKit5D](https://github.com/abcucberkeley/PetaKit5D) style deskewing within the `raymarch.frag` fragment shader for instantaneous feedback. **Shear Angle** is set as a UI parameter, **Pixel Size** is pulled from the Camera configuration, and **Z-Step Size** is based on the stacking step size.

***TIP***: Press **TAB** in the viewer to switch between *Orthogonal* and *Perspective* projection modes. Orthogonal is better for getting MIP-like views to examine the shearing. Hold **SHIFT** to step in 15 degree increments to make this even easier.

#### New: Standalone Volume Viewer (`view/display_backends/volume_viewer_standalone.py`)

A lightweight offline viewer for inspecting TIFF stacks outside of a live experiment. Launch it with:

```bash
navigate -v
# or
navigate --viewer
```

A small Tkinter window opens. Drag and drop one or more `.tif` / `.tiff` files onto it. The viewer reads z-spacing from the TIFF `ImageDescription` tag and XY pixel size from `XResolution` / `ResolutionUnit`, then uploads the stack to the GPU. Multiple files are treated as separate channels and blended together. Each channel gets its own colour, min/max, and autoscale controls.

---

### Reviewer Testing Checklist

We'd love hardware coverage across different GPUs and platforms. Here's the suggested sequence:

**Step 1 — Verify the guard works (no OpenGL install needed)**

1. Check out this branch.
2. Do **not** install `[opengl]`.
3. Run navigate normally through a full acquisition.
4. Confirm the Volume Rendering panel does not appear and nothing is broken.

**Step 2 — Test the live renderer**

1. `pip install -e .[opengl]`
2. Run navigate and open the Camera View.
3. Check the "Enable" checkbox in the Volume Rendering panel — a GLFW window should appear.
4. Run a z-stack acquisition and watch the volume build up in real time.
5. Adjust Opacity and World Step and verify the render updates live.
6. **Try different Color channels in Single and Overlay Mode** --> This will likely be a source of bugs.

**Step 3 — Test the standalone viewer**

1. Run `navigate -v`
2. Drag a multi-slice TIFF file onto the window.
3. Verify the volume renders and controls respond.
4. Try dragging a second file — old channel widgets should be replaced cleanly.

**Also welcome:** bug reports, code structure feedback, any crashes or GL errors in the log.


src/navigate/view/display_backends/volume_viewer_standalone.py
```diff
@@ -0,0 +1,360 @@
...
+            return
+
+        self.parent.backend.set_min_max([_min, _max], self.id)
+
+    def load_stack(self, stack_path: str) -> np.ndarray:
+        with tifffile.TiffFile(stack_path) as tif:
+            
+            # z-spacing
+            try:
+                image_desc = dict(eval(tif.pages[0].tags['ImageDescription'].value))
```
AdvancedImagingUTSW, line 247: **[P0] Do not evaluate TIFF metadata.** `ImageDescription` is controlled by the dropped file, so `eval()` permits arbitrary Python execution before this exception handler can fall back. Please use tifffile's parsed metadata, JSON, `ast.literal_eval` with strict schema validation, or another non-executing parser, and add a malicious-metadata regression test.

src/navigate/main.py
```diff
@@ -41,6 +41,7 @@
 from navigate.controller.controller import Controller
 from navigate.controller.configurator import Configurator
 from navigate.log_files.log_functions import log_setup
+from navigate.view.display_backends.volume_viewer_standalone import VolumeViewer
```
AdvancedImagingUTSW: **[P1] Keep the OpenGL extra genuinely optional.** This module-level import immediately loads `tkinterdnd2`, GLFW, PyGLM, and PyOpenGL. A normal `.[dev]` installation therefore cannot even import `navigate.main`; this is the direct cause of the current Windows CI collection failure. Move the complete standalone-viewer import into the `--viewer` branch and handle failure of the full optional stack there.

src/navigate/view/display_backends/gl_backend.py
```diff
@@ -0,0 +1,1204 @@
...
+
+        # Reset flag
+        self.need_render = False
+
+    def _init_gl_resources(self):
+        """Initialize shaders and set uniforms, create VAO, and set up camera."""
+
+        import os
+        from pathlib import Path
+        shader_dir = os.path.join(Path(__file__).parent, "shaders")
```
AdvancedImagingUTSW, line 949: **[P1] These runtime shader files are absent from built wheels.** I built a wheel from this head and confirmed it contains the Python backend files but none of `shaders/*.vert` or `shaders/*.frag`. Declare the resources as package data and load them through `importlib.resources`; add a wheel-install smoke test so editable installs cannot mask this.

src/navigate/view/display_backends/gl_backend.py
```diff
@@ -0,0 +1,1204 @@
...
+
+        if not (window and vao and shader):
+            raise RuntimeError("GL not ready yet")
+
+    def _render_thread(self, window_dim: tuple):
+        global GL
+
+        try:
+            # Initialize GLFW
+            if not glfw.init():
```
AdvancedImagingUTSW, line 754: **[P1] This GLFW lifecycle is not cross-platform.** GLFW initialization, window/context creation, event polling, window destruction, and termination are main-thread-only APIs. They are all run on this worker thread. Separately, the 4.3 context request and GLSL 4.30 shaders cannot run on macOS, whose OpenGL support tops out at 4.1, and they contradict the PR's OpenGL 3.3+ requirement. The window/event architecture and advertised GL baseline both need redesign.

src/navigate/view/display_backends/gl_backend.py
```diff
@@ -0,0 +1,1204 @@
...
+            # signal that we're ready to receive commands and data
+            self.is_ready.set()  
+
+            # run the render loop
+            self._render_loop()
+
+        except Exception as e:
+            print("[GL Thread] Fatal Error:", e)
+            traceback.print_exc()
+            self.is_ready.set()  # unblock start() if waiting
```
AdvancedImagingUTSW, line 792: **[P1] Do not signal successful readiness on fatal initialization.** `start()` returns after this event, but `is_running` remains set even though the thread exits. `thread_is_running()` then lies, `stop()` returns early without clearing the event, and acquisition continues placing images into an unbounded queue. Propagate startup success/failure explicitly, clear lifecycle state in `finally`, bound/drop stale frames, and test failed initialization plus restart.

src/navigate/view/display_backends/volume_viewer_standalone.py
```diff
@@ -0,0 +1,360 @@
...
+                input_var=tk.DoubleVar(value=45.0),
+                input_args={"from_": -90.0, "to": 90.0, "increment": 0.5, "width": 8},
+                label_pos="top",
+            ),
+            "dz": LabelInput(
+                parent=volume_settings_frame,
+                label="Z-Step Size",
+                input_class=ValidatedSpinbox,
+                input_var=tk.DoubleVar(value=0.4),
+                input_args={"from_": 0.01, "to": np.Inf, "increment": 0.01, "width": 8},
```
AdvancedImagingUTSW, line 134: **[P1] `np.Inf` was removed in NumPy 2.0.** The current CI dependency resolution installs NumPy 2.0.2, so constructing the standalone viewer will raise here after the optional dependencies are present. Use `np.inf` and add a construction smoke test under the project's resolved dependency set.

src/navigate/view/display_backends/volume_viewer_standalone.py
```diff
@@ -0,0 +1,360 @@
...
+        try:
+            value = float(variable.get())
+        except:
+            # Handle "" and other such invalid inputs
+            return
+        
+        exec(f"self.backend.set_{field}({value})")
+
+    def on_drop(self, event):
+        dropped_files = event.data.split()
```
AdvancedImagingUTSW, line 321: **[P2] Preserve drag-and-drop paths containing spaces.** TkDND supplies a Tcl list and commonly brace-quotes Windows paths. Plain string splitting breaks those paths into multiple entries. Use `self.view.tk.splitlist(event.data)` and validate the resulting paths/channel count.

src/navigate/model/devices/APIs/imagineoptics/imop.py
```diff
@@ -775,7 +775,7 @@ def __init__(
             basepath, "MirrorFiles", "FlouresceinOctober12.wcs"
         ),
         interaction_matrix_file_path=os.path.join(
-            basepath, "MirrorFiles", "OlympusApril22.aoc"
+            basepath, "MirrorFiles", "InfluenceMatrix-20260527_Pupil18mm.aoc"
```
AdvancedImagingUTSW, line 778: **[P2] Remove this unrelated calibration-default change from the OpenGL PR.** The new site/date-specific `.aoc` file is not included here, and changing this default can silently break existing Imagine Optics setups that provide the previous external calibration file. Please revert it or submit it separately with its hardware rationale and deployment plan.

AdvancedImagingUTSW (changes requested):
## Review result: Request changes

This is a substantial and promising addition, but the current implementation is not ready to merge. It does not yet meet the stated backwards-compatibility or macOS/Linux/Windows requirements, and it contains a security-critical TIFF metadata issue.

### Release-blocking findings

1. **Arbitrary code execution from TIFF metadata:** the standalone viewer evaluates user-controlled `ImageDescription` text with `eval()`.
2. **The optional dependency guard is ineffective:** importing `navigate.main` unconditionally imports the standalone viewer and its optional dependencies. The current Windows Actions failure and a clean local reproduction both stop during test collection with `ModuleNotFoundError: tkinterdnd2`.
3. **Built wheels omit the GLSL shaders:** the renderer opens four shader files at runtime, but the package does not declare them as package data. A wheel built from this head contained the Python backend files but none of the `.vert` or `.frag` resources.
4. **The renderer cannot support macOS as written:** GLFW initialization, window creation, event polling, and destruction occur on a worker thread even though GLFW requires these operations on the main thread. The backend also requests OpenGL 4.3 and uses GLSL 4.30, while macOS supports at most OpenGL 4.1. This also contradicts the advertised OpenGL 3.3+ minimum.
5. **Fatal renderer startup is treated as success:** the exception path signals readiness but leaves `is_running` set. The thread can be dead while `thread_is_running()` returns true, after which acquisition images can accumulate in the unbounded queue.
6. **The standalone viewer is incompatible with the dependency versions CI currently resolves:** it uses `np.Inf`, which was removed in NumPy 2.0.

### Additional issues

- Drag-and-drop parsing uses `event.data.split()`, which breaks paths containing spaces and Tcl brace quoting.
- The PR includes an unrelated change to the default Imagine Optics calibration filename without including the referenced file.
- No tests or documentation files were added for approximately 2,100 lines of new behavior.
- The CI matrix is Windows-only, so there is no macOS or Linux validation.
- `black --check` would reformat six changed Python files, and `git diff --check` reports extensive trailing whitespace.

### Validation performed

- Reviewed current head `defbacfc4b14cecfa5433b37f46081987b226640`.
- Confirmed the current Windows Actions run fails during collection because the optional dependency is imported unconditionally.
- Reproduced the same two collection errors locally without the OpenGL extra.
- Installed the optional dependencies into an isolated temporary target and reran the focused main tests: 2 passed and 2 failed because the existing main tests were not updated for the new `viewer` branch.
- Built and inspected the wheel, confirming that the shader resources are absent.
- Reproduced the failed-start lifecycle state where `is_running=True` while the render thread is no longer alive.

Please address the security issue and all release-blocking findings, add focused unit/package tests plus macOS/Linux/Windows CI coverage, and rerun the complete suite before requesting another review.

References: [GLFW thread-safety requirements](https://www.glfw.org/docs/latest/intro), [GLFW macOS OpenGL compatibility](https://www.glfw.org/docs/latest/compat_guide.html), [NumPy 2.0 removals](https://numpy.org/doc/2.1/release/2.0.0-notes.html).

src/navigate/main.py
AdvancedImagingUTSW: Addressed in `a6c33a892`. The standalone viewer is now imported only when `--viewer` is requested. Missing optional packages produce beginner-oriented terminal guidance using the active Python interpreter, while the technical traceback is logged and startup resources are cleaned up safely. Regression coverage blocks `OpenGL`, `glfw`, `glm`, and `tkinterdnd2` during `import navigate.main`; 13 focused main/parser tests, direct import isolation, compileall, Black, and whitespace checks pass. The new Windows Actions run is queued.

AdvancedImagingUTSW:
I believe we will proceed with @conorhughmcfadden's standalone PyPi version.

PR #1229, opened 2026-07-21
Random fix: handle missing camera config values and persist channel setting update

annie-xd-wang:
This PR adds a few small stability fixes across configuration, camera, channel, and common utility code.

        camera_settings.py: avoid crashing when pixel size is missing for the selected microscope/zoom and log a warning instead
        channels_settings.py: write updated channel settings back to channel_setting_dict after widget changes
        common_functions.py: make copy_proxy_object() recursively copy dict- and list-like objects more generally
        configuration_controller.py: clean up the missing-microscope warning message formatting

### DIFF
--- a/src/navigate/controller/configuration_controller.py
+++ b/src/navigate/controller/configuration_controller.py
@@ -99,7 +99,10 @@ class ConfigurationController:
             microscope_name = self.configuration["experiment"]["MicroscopeState"][
                 "microscope_name"
             ]
-        if microscope_name not in self.configuration["configuration"]["microscopes"].keys():
+        if (
+            microscope_name
+            not in self.configuration["configuration"]["microscopes"].keys()
+        ):
             logger.warning(
                 f"Microscope {microscope_name} not found in configuration. Available microscopes: "
                 f" {list(self.configuration['configuration']['microscopes'].keys())}"
@@ -148,7 +151,9 @@ class ConfigurationController:
         setting = {
             "laser": self.lasers_info,
         }
-        for i, filter_wheel_config in enumerate(self.microscope_config.get("filter_wheel", [])):
+        for i, filter_wheel_config in enumerate(
+            self.microscope_config.get("filter_wheel", [])
+        ):
             filter_wheel_name = filter_wheel_config.get("name", f"FilterWheel-{i}")
             setting[filter_wheel_name] = list(
                 filter_wheel_config["available_filters"].keys()
@@ -624,6 +629,25 @@ class ConfigurationController:
         return self.configuration["waveform_constants"]["remote_focus_constants"][
             microscope_name
         ].keys()
+
+    def get_zoom_pixel_sizes(self, microscope_name: str) -> dict:
+        """Return a dictionary of pixel sizes
+
+        Returns
+        -------
+        pixel_size_dict : dict
+            A dictionary of pixel sizes: {zoom_value : pixel_size}
+        """
+        if (
+            microscope_name is None
+            or microscope_name
+            not in self.configuration["configuration"]["microscopes"].keys()
+        ):
+            microscope_name = self.microscope_name
+
+        return self.configuration["configuration"]["microscopes"][microscope_name][
+            "zoom"
+        ]["pixel_size"].copy()
 
     @property
     def gui_setting(self) -> dict:
--- a/src/navigate/controller/configuration_controller.py
+++ b/src/navigate/controller/configuration_controller.py
@@ -99,6 +99,12 @@ class ConfigurationController:
             microscope_name = self.configuration["experiment"]["MicroscopeState"][
                 "microscope_name"
             ]
+        if microscope_name not in self.configuration["configuration"]["microscopes"].keys():
+            logger.warning(
+                f"Microscope {microscope_name} not found in configuration. Available microscopes: "
+                f" {list(self.configuration['configuration']['microscopes'].keys())}"
+            )
+            return False
 
         assert (
             microscope_name in self.configuration["configuration"]["microscopes"].keys()
--- a/src/navigate/controller/sub_controllers/camera_settings.py
+++ b/src/navigate/controller/sub_controllers/camera_settings.py
@@ -361,9 +361,6 @@ class CameraSettingController(GUIController):
         #: str: Mode value
         self.mode = "stop"
 
-        #: str: Solvent type
-        self.solvent = "BABB"
-
         # Getting Widgets/Buttons
 
         #: dict: Mode widgets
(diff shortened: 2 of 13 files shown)


src/navigate/controller/sub_controllers/camera_settings.py
```diff
@@ -859,9 +859,15 @@ def calculate_physical_dimensions(self):
...
-        ][microscope_name]["zoom"]["pixel_size"][zoom]
+        try:
+            pixel_size = self.parent_controller.configuration["configuration"][
+                "microscopes"
+            ][microscope_name]["zoom"]["pixel_size"][zoom]
+        except KeyError:
+            logger.warning(
+                f"Pixel size for microscope {microscope_name} and zoom {zoom} not found."
+            )
+            return
```
AdvancedImagingUTSW: [P1] Do not leave a stale calibrated FOV active

When this lookup fails, returning here leaves `FOV_X` and `FOV_Y` at their values from the previous microscope/zoom; the added test explicitly asserts that behavior. The tiling controller later consumes those widgets as physical stage spacing, so switching to an uncalibrated zoom can silently produce incorrect tile coordinates instead of merely skipping a display update. Please invalidate or disable the FOV values, or propagate failure so tiling/acquisition cannot proceed until calibration exists, and update the regression test accordingly.

src/navigate/controller/configuration_controller.py
```diff
@@ -99,6 +99,12 @@ def change_microscope(self, microscope_name=None) -> bool:
             microscope_name = self.configuration["experiment"]["MicroscopeState"][
                 "microscope_name"
             ]
+        if microscope_name not in self.configuration["configuration"]["microscopes"].keys():
+            logger.warning(
+                f"Microscope {microscope_name} not found in configuration. Available microscopes: "
+                f" {list(self.configuration['configuration']['microscopes'].keys())}"
+            )
+            return False
```
AdvancedImagingUTSW, line 110: [P2] Restore experiment state when the microscope is rejected

`Controller.change_microscope()` writes the requested name into `experiment["MicroscopeState"]` before calling this method. Returning `False` here therefore leaves the experiment dictionary holding the missing name while this controller retains the previous microscope and configuration. Downstream code such as the tiling controller indexes the hardware configuration with the experiment name and can then raise `KeyError`. Please validate before mutating the experiment state or restore the previous name on failure, with a regression test through the outer controller path.

AdvancedImagingUTSW (changes requested):
Requesting changes for two correctness issues in the new fallback paths. The focused and surrounding local suites pass (126 tests total), but the missing-calibration path preserves stale physical dimensions, and the missing-microscope path can leave experiment and controller state inconsistent.

src/navigate/controller/sub_controllers/camera_settings.py
annie-xd-wang: calculate_physical_dimensions() now returns True/False to indicate whether the physical-dimension calculation succeeded, and that failure is surfaced to the user through a GUI warning before acquisition starts.
The camera settings controller already normalizes the ROI width and height to valid values before acquisition, so stale widget values are not the source of failure here. The remaining failure mode is a missing or invalid pixel_size entry in configuration.yaml. When that happens, the calculation fails explicitly and the user gets a clear warning telling them to check the microscope/zoom calibration in the configuration file.

src/navigate/controller/configuration_controller.py
annie-xd-wang: The function has been updated to validate the microscope_name and zoom value before applying any changes. The controller now falls back to a valid setting when the requested microscope/zoom combination is invalid, and a warning is shown to the user.

AdvancedImagingUTSW:
Hi @annie-xd-wang — I pushed five follow-up commits based on the current PR head (`b7a3bb5`) to [`kdean/fix-null-pixel-size`](https://github.com/TheDeanLab/navigate/tree/kdean/fix-null-pixel-size). These commits are on the follow-up branch and are not yet part of this PR head.

Commit-by-commit summary:

1. [`e388c1f68`](https://github.com/TheDeanLab/navigate/commit/e388c1f683e030feadf50108a52d4f8e3476c84a) — **Handle YAML-null pixel sizes**
   - **Problem:** YAML `null` becomes Python `None`, so `float(None)` raised `TypeError`; the existing handlers only caught `KeyError`/`ValueError`.
   - **Approach:** Catch `TypeError` in both primary- and additional-microscope pixel-size parsing paths and add a regression test for the warning/failure path.

2. [`ff108b6c3`](https://github.com/TheDeanLab/navigate/commit/ff108b6c3041a31db252ad80a4b214c5236d84b0) — **Align Configure Microscopes controls with their labels**
   - **Problem:** The Zoom Value combobox was placed on the Setting row, and the Setting combobox was placed on an unlabeled row below it.
   - **Approach:** Correct the two grid-row indices and add a layout regression test.

3. [`009183334`](https://github.com/TheDeanLab/navigate/commit/009183334b95b379f69530d261b97594f1faf529) — **Expand microscope-popup tests**
   - **Problem:** The popup had little coverage beyond the row-alignment regression.
   - **Approach:** Add multi-microscope tests for device-label deduplication, configured/missing values, disabled/readonly states, column layout, getters, variables, and buttons.

4. [`8d6ebb524`](https://github.com/TheDeanLab/navigate/commit/8d6ebb52447b62d4f760832766270e6231d273fc) — **Preserve additional-microscope calibration**
   - **Problem:** Before acquisition, `update_experiment_values()` replaced an additional microscope's zoom calibration (for example, `9.7`) with the camera sensor pitch (`6.5`).
   - **Approach:** Restrict the sensor-pitch assignment to the primary camera-settings controller, preserving the selected calibration for additional microscopes; add a regression that verifies both the retained calibration and the resulting FOV.

5. [`bdca4cbae`](https://github.com/TheDeanLab/navigate/commit/bdca4cbaebaedfe9a9d0f2517ccb66c727b7279d) — **Invalidate stale FOV values in the tiling workflow**
   - **Problem:** Failed physical-dimension calculation left old camera FOV values in place, and an open Tiling Wizard could retain its copied values and generate positions from stale spacing.
   - **Approach:** Clear invalid camera FOV values, invalidate copied X/Y tiling FOV state, revalidate before opening the wizard and before populating the table, and fail closed with a warning. The validation warning now reads: “Can't calculate positions, please make sure the camera pixel size is correct.”

Validation completed on the follow-up branch:

- Camera settings: **36 passed**
- Tiling controller: **20 passed**
- Channels tab: **13 passed**
- Popup views: **18 passed**

Could you please test this branch once more before we merge—especially selecting an additional microscope and zoom, confirming its calibrated pixel size/FOV survives acquisition preparation, and exercising the Tiling Wizard with valid, missing, and YAML-null pixel sizes? If that behavior looks correct in the application, we can then bring these commits onto the PR branch for final CI and merge review.

annie-xd-wang:
Confirmed that the branch [kdean/fix-null-pixel-size](https://github.com/TheDeanLab/navigate/tree/kdean/fix-null-pixel-size) works as expected.

AdvancedImagingUTSW (approved):

PR #1230, opened 2026-07-22
Replace ImagineOptics ctypes wrapper with wavekit_py SDK

DUMUduwaka:
### Changes

- Replaces the ctypes wrapper around `c_interface_vc100_x64.dll `with the official `wavekit_py` package.

- Removes the Pointer base class and the private `__Imop_ *` methods that wrapped individual C API calls. Mirror functionality is now exposed through Python classes (HasoSlopes, Pupil, ModalCoef, CorrDataManager, WavefrontCorrector).

- Replaces silent failures with actionable errors: `ImportError` with installer instructions when `wavekit_py` is missing, `RuntimeError` when the wavefront-corrector .dat config can't be read, and `ConnectionError` when the mirror can't be reached (covering power, USB, drivers, and another program holding the connection).

- Adds E_ZERNIKE_INDEXING_T (WYANT/NOLL) to `enums.py`.

- Updates the mirror piston plot in `controller/sub_controllers/adaptive_optics.py`: the color range was fixed at ±0.25, which clipped or washed out the display depending on wavefront amplitude. It now scales to the data via `np.nanmax(np.abs(img))`, with a colorbar that updates in place.

Note:
The SDK base path in `APIs/ImagineOptics/imop.py` is currently hardcoded to a machine-specific location.


src/navigate/model/devices/APIs/imagineoptics/imop.py
```diff
@@ -1,16 +1,32 @@
...
+    E_MODAL_T,
+    E_PUPIL_DETECTION_T,
+    E_ZERNIKE_NORM_T,
+    E_PUPIL_COVERING_T,
+    E_ZERNIKE_INDEXING_T,
+)
 
-basepath = "D:\\WaveKitX64"
+# Define the base path where the mirror files and wfc and wfs configuration files
+basepath = "C:/Users/Spectral/Desktop/muDM"
```
AdvancedImagingUTSW: [P1] This hard-coded contributor-specific muDM path makes the driver unusable on every other machine and replaces the documented Mirao 52E calibration with a different device's files. `ImagineOpticsMirror.connect()` still accepts no connection parameters, so users cannot override the WFC, HASO, flat, or interaction-matrix paths before construction. Please expose these paths through the microscope hardware configuration and `get_connect_params()` instead of shipping universal machine-specific defaults.

src/navigate/model/devices/APIs/imagineoptics/imop.py
```diff
@@ -47,224 +63,244 @@
...
         pupil.set_data(pupil_buffer)
 
-        self.pupil_center, self.pupil_radius = PupilCompute().fit_zernike_pupil(
-            pupil,
+        zernike_pupil = wk.ComputePupil.fit_zernike_pupil(
+            pupil.pointer,
             E_PUPIL_DETECTION_T.E_PUPIL_AUTOMATIC,
-            E_PUPIL_COVERING_T.E_PUPIL_CIRCUMSCRIBED,
-            0,
+            E_PUPIL_COVERING_T.E_PUPIL_INSCRIBED,
```
AdvancedImagingUTSW: [P1] This migration changes the fitted Zernike pupil from `E_PUPIL_CIRCUMSCRIBED` to `E_PUPIL_INSCRIBED`. That changes the fitted radius and modal basis, so existing calibrated systems can generate different mirror commands even though this PR is presented as an SDK migration. Unless this is an intentional device-specific recalibration backed by hardware results, preserve the old covering convention or make it explicit in configuration.

src/navigate/model/devices/APIs/imagineoptics/imop.py
```diff
@@ -275,747 +311,501 @@ def load_wcs(self, path=None, name=None, mode_file=False):
...
+            zernike_indexing=zernike_indexing,
+            projection_pupil=projection_pupil,
+        )
+
+    def set_data(self, coef, pupil=None):
+        """
+        Set polynomial coefficients information and values.
+        """
+        pupil_obj = pupil.pointer if pupil is not None else None
+        self._modal_coef.set_data(np.asarray(coef).astype(np.float32), pupil_obj)
```
AdvancedImagingUTSW, line 514: [P1 risk] The old implementation explicitly supplied coefficient indices `1..n_modes`; this call removes them. The available WaveKit 4.3 Python API expects `set_data(coef_array, index_array, pupil)`, while the target 4.5.1 SDK contract is not present in the repository. Please add a contract test based on the exact 4.5.1 signature and prove that indices are preserved. If the signature did not change, `display_modes()` catches the resulting exception and returns without moving or flattening the mirror.

AdvancedImagingUTSW (changes requested):
Requesting changes for two confirmed blockers and one high-risk SDK-contract gap.

Validation performed:
- The five existing ImagineOptics tests pass in the project `navigate` Conda environment, but they replace the entire low-level API module with a fake and execute none of this migration.
- Compilation, Black, and `git diff --check` pass.
- GitHub Actions run 29884864573 ended `action_required` with zero jobs, so CI has not executed.
- No tests or documentation were added for the 496-line SDK integration.
- The proprietary WaveKit 4.5.1 SDK is not available in the repository or local review environment, so its exact call contract and hardware behavior remain unverified.

Please address the inline blockers, add SDK-contract tests that exercise this module rather than mocking it wholesale, update the ImagineOptics configuration documentation, and rerun CI before merge.

DUMUduwaka:
**Address review: configurable SDK paths, pupil covering revert, SDK-contract tests**

- Require wfc/haso/positions/interaction-matrix config paths via
  get_connect_params() instead of hardcoding the muDM desktop path;
  expose them in configuration_database.py and document them.

- Revert Zernike pupil covering to E_PUPIL_CIRCUMSCRIBED (the earlier
  switch to E_PUPIL_INSCRIBED was unintentional)

- Add SDK-contract tests that inspect the real wavekit_py signatures
  (skipped when the SDK isn't installed).

src/navigate/model/devices/APIs/imagineoptics/imop.py
```diff
@@ -47,224 +66,240 @@
...
-        self.pupil_center, self.pupil_radius = PupilCompute().fit_zernike_pupil(
-            pupil,
+        zernike_pupil = wk.ComputePupil.fit_zernike_pupil(
+            pupil.pointer,
             E_PUPIL_DETECTION_T.E_PUPIL_AUTOMATIC,
             E_PUPIL_COVERING_T.E_PUPIL_CIRCUMSCRIBED,
-            0,
+            False,
         )
+        self.pupil_center = zernike_pupil.center
```
AdvancedImagingUTSW, line 152: [P1] The added contract test checks only the four input parameters, but this call's return contract is equally critical. Current WaveKit usage unpacks `fit_zernike_pupil(...)` as `(center, radius)`; this code instead dereferences `.center` and `.radius`, which would abort mirror initialization if WaveKit 4.5.1 still returns a tuple. Please run this against the target SDK and either unpack the tuple or add a real-SDK assertion that proves the object return shape.

AdvancedImagingUTSW (changes requested):
Re-reviewing the updated head `f86a719`. The configurable calibration paths and restoration of the circumscribed pupil convention address two earlier concerns, but I am still requesting changes before merge.

Remaining blockers and regressions:
- The target WaveKit 4.5.1 contract is still unverified. The new contract module skips when the proprietary SDK is unavailable, and it did not run in this review environment.
- `ModalCoef.set_data()` still omits the coefficient-index array. If 4.5.1 retains that argument, `display_modes()` catches the resulting exception and returns without moving or flattening the mirror.
- `fit_zernike_pupil()` is treated as returning an object with `.center` and `.radius`, but the new test checks only its input signature, not its return shape. See the inline comment.
- `load_wcs(name=...)` and `save_wcs(name=...)` still resolve through `C:/Users/Spectral/Desktop/muDM` instead of a configured path.
- `CoreEngine.get_config()` changes the existing `tolerance_radius` key to the misspelled `tolerence_radius`.

Validation:
- 14 focused mirror/factory tests passed; the real-SDK contract module skipped.
- Ruff, Black, and `git diff --check` passed.
- Actions run 29938082325 concluded `action_required` with zero jobs, so CI did not execute.

Before merge, I recommend that @conorhughmcfadden review the WaveKit API assumptions and test this branch on his system with the real SDK and hardware. Please cover initialization, flattening, a representative modal move, and named WCS load/save so we have evidence for both the SDK contract and the feature path.

DUMUduwaka:
### **Thanks for the detailed re-review. Addressed all five items below.**

1. **Verified the WaveKit 4.5.1 SDK contract against a real installation**. Added two tests, one confirming `ModalCoef.set_data()` does not require an index array, and a return-shape test for `fit_zernike_pupil()`. This confirmed that `ModalCoef.set_data()` and `ComputePupil.fit_zernike_pupil()` already matched the installed SDK. 

2. **ModalCoef.set_data() coefficient-index array**. Checked directly against the installed 4.5.1 source. `set_data(self, coef_array, pupil)` takes 2 arguments, no index array in this version. `imop.py`'s call already matches. `test_set_data_does_not_require_an_explicit_index_array` asserts this explicitly.

3. **fit_zernike_pupil() return shape**. Also confirmed against the installed SDK. It returns `ZernikePupil_t`, a `namedtuple` with fields `('center', 'radius')` in that order, so .center/.radius access is correct. Added `test_fit_zernike_pupil_return_shape,` which asserts this field order directly.

4. **Removed hardcoded `C:/Users/Spectral/Desktop/muDM` path**. `load_wcs(name=...)`/`save_wcs(name=...)` now resolve through `mirror_files_path`, which defaults to the directory containing `flat_path `(already a required config value) and can be overridden with a new optional `hardware/mirror_files_path` config key if someone wants WCS saves/loads to land elsewhere. No machine-specific path remains anywhere in the code.

5. **Fixed `tolerance_radius`**, misspelled `tolerence_radius` during the SDK migration.

### Validation

- SDK-contract tests run against the real, installed WaveKit 4.5.1 SDK: **8/8 passed**.
```
    TestModalCoefSetDataContract::test_set_data_does_not_require_an_explicit_index_array
    TestComputePupilContract::test_fit_zernike_pupil_signature
    TestComputePupilContract::test_fit_zernike_pupil_return_shape
    TestCorrDataManagerContract::test_get_greatest_common_pupil_takes_no_arguments
    TestCorrDataManagerContract::test_constructor_kwargs
    TestModalCoefPrefsContract::test_set_zernike_prefs_kwargs
    TestHasoConfigContract::test_get_config_returns_three_values
    TestWavefrontCorrectorContract::test_preferences_round_trip_kwargs
```
- Full mirror + ImagineOptics API + config tests: **52/52 passed**.

**I still couldn't test the new changes with real hardware, as the muDM isn't available right now, it's in use on another system.**

@conorhughmcfadden, could you test this branch on your system with the real SDK and hardware?

AdvancedImagingUTSW (dismissed):

conorhughmcfadden:
## Testing notes: legacy config file + AO loop performance

While validating this PR against a real Mirao52-e + HASO4 setup, ran into two issues worth recording here.

### 1. Startup crash: `WaveKit metrology functions not availables`

On first run, `IMOP_Mirror.__init__` failed inside `CorrDataManager.new_from_backup_file` with:
```
Exception('IO_Error', b'Error type invalid_argument.\nWaveKit metrology functions not availables.')
```

Root cause: WaveKit's SDK-feature authorization (metrology, adaptive optics, etc.) is embedded **inside the HASO `.dat` config file itself**, not gated by a machine-wide dongle. This is visible in the SDK's own docs (`inline_check.hpp`):

```cpp
imop::haso::ConfigFile cf(config_file_path);
imop::haso::internal_::SoftwareInfo software_info(cf);
if (!software_info.has_sdk_feature(sdk_metrology))
    throw imop::invalid_argument("WaveKit metrology functions not availables.");
```

The `configuration.yaml` on the test machine still pointed `haso_config_file_path` at a leftover file from the old ctypes-era install (`D:\WaveKitX64\MirrorFiles\HASO4_first_7458.dat`, 278 KB, dated 2018) that predates this authorization scheme. The correctly provisioned copy for WaveSuite 4.5.1 lives at `C:\Program Files\Imagine Optic\Configuration Files\WFS_HASO4_first_7458.dat` (1.25 MB, includes the current auth metadata) — confirmed working since it's the same file WaveTune uses for real closed-loop AO on that machine. **Fix was config-only**: repoint `haso_config_file_path` (and `wfc_config_file_path`) at the current Imagine Optic-provisioned files, not a navigate code change.

### 2. AO closed-loop (TonyWilson) frame rate regression

After the config fix, the mirror worked, but TonyWilson ran noticeably slower than on `develop`. Measured from debug logs (gap between consecutive `Turning on laser` events):

| Branch | ms/step |
|---|---|
| `develop` (ctypes) | ~130 ms |
| this PR (`wavekit_py`) | ~306 ms |

The laser pulse/exposure time itself was identical (~36–38 ms) on both branches, so the ~175 ms/step gap was entirely software-side, between frames.

Traced it to `display_modes()` calling `HasoSlopes.new_from_modal_coef(modal_coef, haso_config_file_path)` once per AO step. That call passes the raw `haso_config_file_path` string into the native DLL every time, and per the SDK's own `check_metrology_function_availability()` (above), a raw path with no cached `ConfigFile` object triggers a full re-parse of that file on every call — now a 1.25 MB file, re-read/re-parsed 27+ times per TonyWilson run.

**Fix:** `CorrDataManager.compute_command_from_modalcoef(modal_coef)` computes the same delta command directly from Zernike coefficients in one call, without a `config_file_path` argument — it operates on the `CorrDataManager` object already validated once at construction, so nothing gets re-parsed per step. Replaced the `HasoSlopes`-based two-step path with this single call and removed the now-unused `HasoSlopes` wrapper class. Added an SDK-contract test asserting the installed `wavekit_py` really exposes `compute_command_from_modalcoef` with the expected signature.

Net result: closed-loop AO frame rate back in line with `develop`.

AdvancedImagingUTSW (approved):

PR #1232, opened 2026-07-24
Fix advanced settings overwriting legacy config when custom configuration is loaded

annie-xd-wang:
This PR fixes a bug where saving advanced camera or stage settings could write to the legacy default configuration file instead of the currently loaded configuration file.

However, there is a related broader behavior to revisit: on application exit, we still save files such as experiment, waveform_constants, and related configs back to the legacy default paths.
I’m still evaluating whether all runtime save flows should consistently write back to the currently loaded custom configuration set instead of the legacy defaults. That broader behavior is not changed in this PR.

### DIFF
--- a/src/navigate/controller/sub_controllers/camera_settings.py
+++ b/src/navigate/controller/sub_controllers/camera_settings.py
@@ -39,7 +39,7 @@ from typing import Optional
 
 # Local Imports
 from navigate.controller.sub_controllers.gui import GUIController
-from navigate.config.config import update_config_dict, get_navigate_path
+from navigate.config.config import update_config_dict
 from navigate.tools.file_functions import write_to_yaml
 from navigate.controller.configuration_controller import ConfigurationController
 
@@ -134,7 +134,7 @@ class AdvancedCameraSettingController:
         # Save the updated configuration to a YAML file.
         write_to_yaml(
             content_dict=self.parent_controller.configuration["configuration"],
-            filename=os.path.join(get_navigate_path(), "config", "configuration.yaml"),
+            filename=self.parent_controller.configuration_path,
         )
 
         # Update the configuration controller with the new configuration.
--- a/src/navigate/controller/sub_controllers/camera_settings.py
+++ b/src/navigate/controller/sub_controllers/camera_settings.py
@@ -32,7 +32,6 @@
 
 # Standard Library Imports
 import logging
-import os
 from typing import Optional
 
 # Third Party Imports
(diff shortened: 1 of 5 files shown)


AdvancedImagingUTSW:
Follow-up cleanup is in commit [`41de24d62`](https://github.com/TheDeanLab/navigate/commit/41de24d62cef9eefe077666d1f3738a515c58c21).

This removes the now-unused `os` imports left behind when the advanced-settings save paths stopped using `get_navigate_path()`, and formats the four touched files with the repository-pinned Black 22.10.0. These changes address the Ruff `F401` and Black formatting issues identified during review; they do not change runtime behavior.

Validation: the scoped Ruff check and Black check pass, and 66 relevant controller tests pass.

AdvancedImagingUTSW:
Added commit [`476352a0a`](https://github.com/TheDeanLab/navigate/commit/476352a0a) to fix the Windows test hang.

The model test fixture configured a multiprocessing logging queue without a listener, then used `Queue.empty()` before `join_thread()` during teardown. Because `Queue.empty()` is not reliable across processes, buffered log records could leave the feeder thread blocked indefinitely. The fixture now actively drains the queue with a `QueueListener`, restores normal logging handlers, stops the listener, and only then closes and joins the queue. The feature-record test now also writes under `tmp_path` instead of the user Navigate directory, preventing local state from leaking into the test.

Validation: `test/model/test_model.py` (12 passed), affected controller tests (66 passed), pinned Ruff 0.0.191, pinned Black 22.10.0, and `git diff --check`.

AdvancedImagingUTSW:
Follow-up commit [`2806b0bb5`](https://github.com/TheDeanLab/navigate/commit/2806b0bb5) fixes the three controller failures revealed once the Windows suite could complete.

The new advanced-stage test had permanently replaced `execute` on the package-scoped dummy controller, so later autofocus and camera-view tests received an empty command list. Both advanced-settings tests now use `monkeypatch` for controller methods, paths, and mutated camera parameters, ensuring all shared state is restored after each test.

Validation: the exact previously failing order passes (4 passed), the full affected controller scope passes (51 passed, 2 skipped), and pinned Ruff 0.0.191, Black 22.10.0, and `git diff --check` pass.

AdvancedImagingUTSW:
Added commit [`039d76faa`](https://github.com/TheDeanLab/navigate/commit/039d76faa) after the replacement Windows run exposed one remaining shared-state issue.

The camera settings live in a `multiprocessing.Manager().dict()` proxy. Using pytest `monkeypatch.setitem` for keys that were originally absent serialized pytest’s internal missing-value sentinel and restored it as a real `<notset>` value. The test now snapshots the proxied camera parameters and explicitly clears/restores them during fixture teardown.

Validation: the exact failing order passes (2 passed), the full affected controller scope passes (89 passed, 2 skipped), and pinned Ruff 0.0.191, Black 22.10.0, and `git diff --check` pass.

AdvancedImagingUTSW (approved):

PR #1234, opened 2026-08-03
Highlight active multiposition row during Z-stack acquisition

annie-xd-wang:
## Summary

- Add a multi-position controller event to select and highlight a table row by its zero-based position index.
- Update Z-stack acquisition to highlight the active position when acquiring across multiple positions.
- Add controller tests for valid and out-of-range row selection.

### DIFF
--- a/src/navigate/controller/sub_controllers/multiposition.py
+++ b/src/navigate/controller/sub_controllers/multiposition.py
@@ -312,6 +312,23 @@ class MultiPositionController(GUIController):
         """
         return self.table.model.df.shape[0]
 
+    def highlight_position(self, row_id: int) -> None:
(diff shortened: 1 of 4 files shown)


AdvancedImagingUTSW:
I like this addition—it makes multi-position Z-stack progress much easier to follow.

I pushed a small follow-up for two reasons:

- The selected row now uses `movetoSelection`, so the active position is scrolled into view instead of being highlighted off-screen when the table contains more rows than the visible viewport.
- The internal event name is corrected from `hightlight_position` to `highlight_position`, with regression coverage for model emission and controller registration so the producer/listener contract stays synchronized.

Validation: the affected controller and Z-stack suite passes (26 tests), and Ruff and diff-integrity checks pass.

Issue #486, opened 2023-06-15
Stop Stages button does not work during ChangeResolution


zacsimile:
We need to be able to stop the stage during `ChangeResolution` in case of a mistake.

zacsimile:
Presumably launching `ChangeResolution` in a thread will solve this? The GUI thread seems to pause when switching between microscopes. It does not pause in the case we are switching resolution within a single microscope.

PR #1237, opened 2026-08-04
Add safe recovery choices after resolution cancellation

AdvancedImagingUTSW:
## Summary

This stacked PR adds the explicit recovery decision requested for a cancelled resolution change:

- show a modal `Resolution Change Cancelled` dialog only after the safety stop has quiesced the resolution worker;
- make Close and Escape select **Keep Current Position** without further stage motion;
- enable **Return to Previous Position** only when the literal pre-movement coordinates are complete, the stopped position was read successfully, all stop attempts succeeded, and every saved axis passes the active stage driver's strict limit validation;
- run an approved return as a separate model-owned task that remains cancellable through **Stop Stage**;
- disable ordinary stage movement controls during the return while leaving the separate Stop Stage control available;
- always restore stage controls on successful, cancelled, failed, or stale return attempts.

The core collision-prevention fix is intentionally isolated in draft PR #1236. This PR is based on that branch so reviewers can test and merge the safety stop independently from the optional recovery UX.

Related to #486.

## Safety behavior

The saved position is copied into a private recovery snapshot tied to the cancelled task and selected microscope. A new resolution task invalidates any older snapshot. Return uses the existing cancellation-aware `Model.move_stage(..., wait_until_done=True)` path and the existing Stop Stage lifecycle; it does not bypass device limit checks or create a second stage executor.

The review pass also caught and fixed a stale-return edge case: if the saved position becomes unavailable between opening the dialog and selecting Return, the model refuses motion, emits a terminal failure event, and the controller re-enables ordinary stage controls.

## Reuse analysis

- Extends the existing model task, event queue, `Model.move_stage`, strict stage `verify_abs_position`, `PopUp`, controller thread pool, and Stop Stage contracts.
- Does not add a recovery-specific device owner, stage limit implementation, movement executor, or stop path.
- Keeps task coordination private to the model and recovery UI coordination private to the controller.
- Design record: [`docs/superpowers/specs/2026-08-04-issue-486-resolution-change-cancellation-design.md`](https://github.com/TheDeanLab/navigate/blob/kdean/issue-486-resolution-recovery/docs/superpowers/specs/2026-08-04-issue-486-resolution-change-cancellation-design.md)
- Executable plan and reuse record: [`docs/superpowers/plans/2026-08-04-issue-486-resolution-change-cancellation.md`](https://github.com/TheDeanLab/navigate/blob/kdean/issue-486-resolution-recovery/docs/superpowers/plans/2026-08-04-issue-486-resolution-change-cancellation.md)

## Validation

- 77 focused recovery, cancellation, controller, popup, feature, and both Stop Stage button tests passed on the rebased head.
- Ruff passed on all Python files changed by this stacked PR.
- The stacked diff whitespace check passed.
- Black passes six of seven changed Python files. It flags one formatting block in model.py that is already present identically on the PR #1236 base and is unrelated to the recovery changes, so this PR intentionally does not reformat it.
- Existing skimage and pandas NumPy deprecation warnings were emitted.

Popup behavior is tested without constructing Tk locally. Although Homebrew Xvfb is installed, Navigate current conda Tk links to macOS AppKit rather than X11, so Xvfb cannot isolate a real window in this environment. Native Windows or Linux X11 Tk under xvfb-run should perform the final constructed-window smoke check.

## Stack

- Base/core safety PR: #1236
- This PR: validated recovery snapshot, cancellable return, and two-choice popup

### DIFF
--- a/docs/superpowers/plans/2026-08-04-issue-486-resolution-change-cancellation.md
+++ b/docs/superpowers/plans/2026-08-04-issue-486-resolution-change-cancellation.md
@@ -601,7 +601,7 @@ git commit -m "feat: add cancellable resolution position recovery" -m \
 - Consumes: `resolution_change_cancelled` and `resolution_return_complete` events; Task 6 recovery command.
 - Produces: `ResolutionChangeCancelledPopup(root, keep_command, return_command, return_enabled)`; controller handlers `_show_resolution_change_cancelled(payload)` and `_finish_resolution_return(payload)`.
 
-- [ ] **Step 1: Write failing popup callback tests without Tk construction**
+- [x] **Step 1: Write failing popup callback tests without Tk construction**
 
 ```python
 def test_popup_close_keeps_current_position():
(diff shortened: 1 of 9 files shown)


AdvancedImagingUTSW:
Hardware testing update: Stop Stage and Return to Previous Position appear to be working. It is not yet perfectly clear whether the microscope configuration is fully correct after stopping, so that aspect would benefit from review and additional verification.

annie-xd-wang (approved):

PR #1240, opened 2026-08-07
Update feature pipeline editor

annie-xd-wang:
## Summary

- Redesign the feature-list editor around a node palette and drag-and-drop board.
- Support inserting, reordering, grouping, and configuring feature nodes, including decision branches and terminal `Break` / `Continue` nodes.
- Add an **Edit Selected Feature List** menu action for custom feature lists, preserving the list name while loading and saving edits.
- Ensure advanced feature settings include a default `None` function option when one was not explicitly configured.
- Add unit coverage for feature graph behavior, grouping validation, menu editing, and default-function persistence.


AdvancedImagingUTSW:
I reviewed the current PR head (`5ec7a2ecf`) and recommend changes before merge. I found three reproducible behavior defects:

1. **[P1] Rightward drag-and-drop can silently preserve the original order.** In `FeatureListGraphController.move_feature`, the code moves both the structure token and the feature, then remaps the structure indices. For a three-node pipeline, dragging the first node into the interior slot before the third node leaves the serialized pipeline unchanged. The reorder logic needs one consistent representation update, with regression coverage for both directions, interior/end positions, and grouped nodes.

2. **[P1] Imported/plugin feature-list edits are exposed but not persisted.** `MenuController.edit_feature_list` allows every non-system list, including Python- and plugin-backed lists. Confirming updates the runtime model, but `FeaturePopupController.update_feature_list` writes YAML only when `module_name` is `None`; those edits therefore disappear after restart. Lists containing external feature classes also cannot be reconstructed by the editor parser/palette. Please either restrict this editor to internally stored lists with a clear message or explicitly convert/save an editable internal copy while preserving the imported source.

3. **[P2] A custom function mapped to `None` suppresses the required default option.** `FeatureAdvancedSettingController.save_parameters` clears `add_default_function_flag` whenever the value field is `None`, even if the function name is not `None`. For example, `{custom_function: None}` is saved without the separate `None: None` option. The flag should be cleared only when the function-name field itself is `None`, with coverage for both cases.

Validation performed on the exact PR tree:

- GitHub Windows CI: `1,787 passed, 13 skipped, 21 deselected`.
- Local focused and nearby consumer tests: 42 passed, with imports confirmed from the isolated PR worktree.
- `git diff --check` passed.
- Pinned Black 22.10.0 would reformat six changed files, and pinned Ruff 0.0.191 reports new line-length violations.

The green suite does not currently exercise the three failing behavior paths above.

AdvancedImagingUTSW:
Implemented and pushed the review fixes in commit `6cf17e206`:

- Reworked feature-node moves around stable old identities, then remapped the feature payloads and grouping-token indexes together. This fixes rightward interior drops and preserves group membership when nodes move into or out of groups.
- Restricted persistent visual editing to Navigate-owned YAML feature lists. The editor now resolves the exact filename from `__sequence.yml`, rejects Python/plugin-owned records at open and Confirm time, saves before changing runtime state, and stays open if saving fails. Pre-acquisition configuration remains runtime-only.
- Normalized advanced-function mappings so every non-empty mapping retains the default `"None": null` entry, even when a custom function also uses a None value.

Validation on the pushed commit: 52 focused controller/conversion tests passed. The repository-wide non-hardware run passed 1,801 tests with 11 skipped and 22 deselected; the one explicit deselection was the OME XML validator because this macOS host has no Java runtime.

Please retest rightward drag-and-drop, moving nodes across group boundaries, persistence after reopening an internal YAML list, rejection of imported/plugin lists, and the advanced-function default option.

AdvancedImagingUTSW:
Rereviewed at `a97bdffbb` and pushed a test-only fix for the failing Windows check.

The failure was in the randomized camera-settings test, not in a pipeline-editor assertion: random 1–3 px light-sheet ROI inputs are clamped by the controller to its 4 px minimum, while the test expected the pre-clamp rounded value of 0. The updated test uses deterministic ordinary and minimum-boundary cases, makes centered-mode setup explicit, and asserts hand-checked dimensions. Local validation passed: 35 camera-settings tests; the full subcontroller suite previously completed with 299 passed and 2 skipped; Black, Ruff, and `git diff --check` are clean. The replacement Windows run is now in progress.

Two implementation risks remain worth watching:

1. The editor round trip still conflates nested-list sequence semantics with tuple loop semantics, and the one-item-group normalization can remove valid singleton nesting after a drag.
2. The custom feature-list `yaml_file_name` read from `__sequence.yml` is not constrained to a basename under the feature-list directory.

I am treating those as follow-up risks rather than blocking this PR. We will merge this and see how it goes, while watching these two areas closely.

